PR #35, opened 2023-01-18
Fix link function syntax in tests. 

yebai:
Some tests still fail, but they are failing on the master branch too.

### DIFF
--- a/test/compiler.jl
+++ b/test/compiler.jl
@@ -11,7 +11,7 @@ using SymbolicPPL:
     ref_to_symbolic,
     addlogicalrules!,
     addstochasticrules!,
-rreshape
+    rreshape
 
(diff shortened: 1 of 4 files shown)


test/compiler.jl
```diff
@@ -10,7 +10,8 @@ using SymbolicPPL:
     ref_to_symbolic!,
     ref_to_symbolic,
     addlogicalrules!,
-    addstochasticrules!
+    addstochasticrules!,
+rreshape
```
yebai, line 14:
```suggestion
    rreshape
```

sunxd3:
Thanks for the fix. I'll go ahead and merge it for now. 

I will open another PR soon to fix all the failed test CI checks.

Issue #202, opened 2024-07-05
Add CI tests against `Tapir` 


yebai:
Related: https://github.com/compintell/Tapir.jl/issues/195 cc @willtebbutt

PR #205, opened 2024-07-25
Make `BUGSModel` parametric so Tapir works with JuliaBUGS models

sunxd3:
Fix https://github.com/TuringLang/JuliaBUGS.jl/issues/202

### DIFF
--- a/test/inference.jl
+++ b/test/inference.jl
@@ -18,14 +18,22 @@ if VERSION >= v"1.10" && VERSION < v"1.11"
         using Pkg
         Pkg.add("Tapir")
         using ADTypes, Tapir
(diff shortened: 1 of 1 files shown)


willtebbutt:
Ugh. Sadly you'll have to do something like [this](https://github.com/TuringLang/Bijectors.jl/blob/7c8b533e4563f5a89b53dddc069e029e655ee370/test/runtests.jl#L37), and only run the Tapir.jl tests if you happen to be testing on version 1.10 or greater.

Sorting this out is on my todo list.

willtebbutt:
Ohhhh ofc, you're completely right 👍

PR #214, opened 2024-09-23
Add some benchmarking code with Stan and Nimble

sunxd3:
The scripts can be used to run examples from Volume 1 of WinBUGS examples.

Stan examples are more optimized (vectorized computation, human effort went into transformed data and generated quantities, etc.). ~~Regardless, on these smaller problems, we should be able to match it's performance when compilation works.~~ With `ReverseDiff` compiled tape, JuliaBUGS gradient and density computation is several times (5- 10x) slower than Stan. `ReverseDiff` is quite optimized, but it still has some fundamental overhead with the tape. To get comparable performance as Stan, I think we need to generate Julia code and use `Mooncake` or `Enzyme`.

Nimble's interface is a bit more tricky, the logp and gradient computation code is from [Chapter 17](https://r-nimble.org/html_manual/example-maximum-likelihood-estimation-using-optim-with-gradients-from-nimderivs..html). (code from [Chapter 16](https://r-nimble.org/html_manual/cha-AD.html#derivatives-involving-model-calculations) runs slowly). I am a bit confused by how to write `nimblefunction`. But the speed is good and it seems to work. Nimble's BUGS examples are a bit out-of-date, it looks to me some examples are from WinBUGS 0.5 era. But the models are generally the same. 

I think moving forward, use Stan as the target is good, easy to setup and also a good performance aim.

- [x] integrate Stan into the current benchmark code


sunxd3:
@yebai a new CI for benchmark is set up and it reports something like https://github.com/TuringLang/JuliaBUGS.jl/pull/214#issuecomment-2392137988. 

I didn't add nimble to the benchmark to be run on PR, but the code is in the repo (we just don't run it), because running Nimble requires more dependency through R etc, and the comparison to Stan should be sufficient information (nimble is 2-3x slower than Stan). 

I am merging this now, can start another PR for further improvement.

PR #219, opened 2024-10-04
Improve graph library

sunxd3:


sunxd3:
This has become a monstrosity -- I have been using it as a test field.
 It will be broken down and introduce into master gradually in the following days.

PR #228, opened 2024-10-24
Improve performance of logdensity computation

sunxd3:


sunxd3:
this PR reduces the logdensity evaluation time of `Rats` from 6ms to about 0.3ms

PR #230, opened 2024-10-27
Mote performance improvement to log density evaluation

sunxd3:
This remove another performance bottleneck: `getindex` from MetaGraph, by storing the data in arrays.


sunxd3:
This is not perfect, but I am merging this now and will optimize memory size later.

PR #233, opened 2024-11-04
Ancestral Sampling and Bayes Ball Algorithm

naseweisssss:


sunxd3:
thanks @naseweisssss, could you resolve the formatter's complaints, also maybe we should have a slightly more complicated model for the test? also, it'll be helpful to run the ancestral sampling multiple time and look at the distributions of the variables.

src/experimental/ProbabilisticGraphicalModels/bayesnet.jl
```diff
@@ -184,11 +199,98 @@ If Z is provided, the conditioning information in `bn` will be ignored.
...
+                    push!(queue, (parent, false))
+                end
+            end
+        end
+    end
+
+    return true
+end
+
+function has_conditioned_descendant(bn::BayesianNetwork, node_id::Int, z_ids::Set{Int})
```
sunxd3, line 275: slight confusion here, why the ball can pass through if the collider child has conditioned descendants? and this could also incur repeated computations?

sunxd3 (commented):
Sorry for the late review, in general this is very good job! Thanks
I left some comments, but all minor.

A suggestion: next time it'll be better to add some descriptions, it doesn't just help me understand the algorithms, but also explain why you might choose some methods over others. You can also factor a lot of these writings in future reports. 

Particularly for implementation of algorithm, a small example would help a lot, giving  some correctness proof (sketches are fine) would be great. Doing this kind of writing is something I didn't do very well at, but it is worthwhile to spent some time on.

src/experimental/ProbabilisticGraphicalModels/bayesnet.jl
```diff
@@ -166,11 +166,26 @@ Ancestral sampling works by:
...
 
-    # TODO: Implement sampling logic
+    for vertex_id in ordered_vertices
+        vertex_name = bn.names[vertex_id]
+        if bn.is_observed[vertex_id]
+            samples[vertex_name] = bn.values[vertex_name]
+            continue
+        end
+        if bn.is_stochastic[vertex_id]
+            dist_idx = findfirst(id -> id == vertex_id, bn.stochastic_ids)
```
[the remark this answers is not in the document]
naseweisssss: If we have `bn.distributions` to have length `nv`, we can eliminate the needs for `findfirst` which might be bottleneck since every pass is a worst case O(N) _I assume thats how Julia works too_. but also increasing memory usage to O(nv) which might not be too much of a problem, compared to the time complexity saved with removing `findfirst`. One thing that I am not sure is how likely that debugging becomes harder if `undef ` are accessed.

src/experimental/ProbabilisticGraphicalModels/bayesnet.jl
```diff
@@ -184,11 +199,98 @@ If Z is provided, the conditioning information in `bn` will be ignored.
...
 
 function is_conditionally_independent(
     bn::BayesianNetwork{V}, X::V, Y::V, Z::Vector{V}
 ) where {V}
-    # TODO: Implement
+    # Get vertex IDs
+    x_id = bn.names_to_ids[X]
+    y_id = bn.names_to_ids[Y]
+    z_ids = Set([bn.names_to_ids[z] for z in Z])
+
```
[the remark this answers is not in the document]
naseweisssss: Added preliminary check and added test case for it

src/experimental/ProbabilisticGraphicalModels/bayesnet.jl
naseweisssss: [https://pmc.ncbi.nlm.nih.gov/articles/PMC6089543/figure/F4/](url)
image

I am considering this. Please let me know what do you think.

naseweisssss: Yes I think it might incur repeated computations, but its a trade off for ensuring all possible paths are considered. I am thinking of keeping a visited nodes list, but unsure if that will cause conflict

naseweisssss: Ah i see makes sense

sunxd3 (approved):

sunxd3:
@naseweisssss thanks a lot for the effort, congrats for the first PR!

PR #259, opened 2024-12-11
Update inprod(a,b)

serenlee:
I found that I have an error with current version inprod(a,b). * is not working for two vectors.


sunxd3:
thanks!

sunxd3:
@serenlee it looks like the formatting is failing but the CI is not working properly, it should be something like

```julia
return LinearAlgebra.dot(a, b)
```

also it would be great if you can add a simple test to file https://github.com/serenlee/JuliaBUGS.jl/blob/patch-1/test/bugs_primitives.jl

sorry for being pedantic, thanks a lot for this

serenlee:
Hi @sunxd3 
Sorry for late response. For unknown reason, I didn’t get notification.

I am actually new to Julia test, - does current test work? Also, your code 
`return LinearAlgebra.dot(a, b) `
is the same as my current one. If you have any feedback, please let me know.

sunxd3:
@serenlee the formatter CI is a bit strange -- it usually show suggestions, but it's not doing that now. I took the liberty to do the formatting, hope it's not offensive.

sunxd3:
@serenlee thanks!

PR #264, opened 2024-12-13
Add permission to benchmark CI

sunxd3:


sunxd3:
okay, the segfault is caused by BridgeStan on macOS

sunxd3:
I am going to merge this now, maybe in some PRs we can disable the MacOS benchmark momentarily

miguelbiron:
Oh yeah that segfault haunted our dreams for a while. It happens only on the newer Github runners with macOS-14 and ARM arch processors. So in Pigeons we decided to just the older runners (macOS-13).

PR #267, opened 2025-01-14
Log Probability Implementation 

naseweisssss:


naseweisssss:
This is the non-dp version for implementation, dp version will be addressed in another PR, still thinking of how to write the test for this

sunxd3:
First, some high-level stuff. Julia has some [conventions on Documentation](https://docs.julialang.org/en/v1/manual/documentation/), sadly, I don't think ChatGPT and Claude understand this yet.

I saw you removed deterministic functions, they do make implementation more difficult, but we can't get rid of them totally if we want to use this for future integration with JuliaBUGS' frontend.

How about try to emulate the implementation [here](https://github.com/TuringLang/JuliaBUGS.jl/blob/79235e97c88a5d19443eb4688dc61519cefe39f2/src/model.jl#L537-L599) to compute the log density, but adapt it to `BayesianNetwork`. 

Dynamic Programming version is in no hurry, we first get the naive recursion correct, then we worry about make DP work.

PR #276, opened 2025-02-14
evaluate function matching BUGSModel

naseweisssss:
This PR should contain evaluate function matching BUGSModel. This version currently have potential problems with ordering of inneighbours order with parameter order. Also assuming all model are transformed.


src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -161,3 +161,35 @@ function add_edge!(bn::BayesianNetwork{V,T}, from::V, to::V)::Bool where {T,V}
...
+            value = fn(parent_vals...)
+            evaluation_env = BangBang.setindex!!(evaluation_env, value, varname)
+            deterministic_index += 1
+        end
+    end
+    return evaluation_env, logp
+end
+
+function parent_values(bn::BayesianNetwork, i::Int)
+    parent_ids = inneighbors(bn.graph, i)
```
sunxd3, line 192: Here you are assuming the argument order is consistent with the order given by inneighbors. 

This won't work with current JuliaBUGS, because node functions in JuliaBUGS have two arguments: evaluation_env and loop_vars.

I wanted to point this out for future ref. It's okay for this PR to work with simpler representations, e.g., the test you wrote.

There is a potential problem with the order, inneighbors as far as I can recall, returns a set. The order of parent's values in turn relies on the interaction order of the set. So the argument order might not be consistent with how the function is defined.

sunxd3 (changes requested):

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
naseweisssss: `    parent_ids = sort(inneighbors(bn.graph, i))`

I am not sure if I understand this right, but do you think does sorting makes the order consistent overall? For example, the code above

sunxd3: consider a node function is `(b, a) -> a + b` (slightly modified your example from test).

sorting might give order `a, b`, but the argument order is `b, a`. So you would need to "remember" the order somehow, or use keyword argument.

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -161,3 +161,42 @@ function add_edge!(bn::BayesianNetwork{V,T}, from::V, to::V)::Bool where {T,V}
...
+    for (i, varname) in enumerate(bn.names)
+        is_stochastic = bn.is_stochastic[i]
+        if is_stochastic
+            dist_fn = bn.distributions[stochastic_index]
+            parent_vals = parent_values(bn, i)
+            dist = dist_fn(parent_vals...)
+
+            value = get(evaluation_env, varname, nothing)
+            bijector = Bijectors.bijector(dist)
+            value_transformed = Bijectors.transform(bijector, value)
```
sunxd3, line 180: I am a bit concerned this might be confusing, I only did this to keep some consistency. In the version of `evaluate` that accept a vector of parameter values, the values are already transformed. Just FYI.

sunxd3 (commented):
In general, it's a solid step. 

But `evaluate` now doesn't work with `BayesianNetwork` translated from `BUGSModel` because you made some assumptions about specific forms of `node function`s in `evaluate` function. This is fine though.

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -10,7 +10,8 @@ struct BayesianNetwork{V,T,F}
     "mapping from variable names to ids"
     names_to_ids::Dict{V,T}
     "values of each variable in the network"
-    values::Dict{V,Any} # TODO: make it a NamedTuple for better performance in the future
+    evaluation_env::NamedTuple
+    loop_vars::NamedTuple
```
sunxd3, line 14: `loop_vars` is different for each variable, so should be a dict.

For instance,
```julia
@bugs begin
    for i in 1:3
        x[i] ~ Normal()
    end
end
```
`x[1]` would have `loop_vars = (i=1,)`, `x[2]` would have `loop_vars = (i=2,)`, etc.

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -29,7 +30,8 @@ function BayesianNetwork{V}() where {V}
         SimpleDiGraph{Int}(), # by default, vertex ids are integers
         V[],
         Dict{V,Int}(),
-        Dict{V,Any}(),
+        (;),    # Empty NamedTuple for evaluation_env
```
sunxd3, line 33: Because `BayesianNetwork` is immutable, although we could use `BangBang.setfield!!` to set it later, maybe a better way to define this constructor is to take `evaluation_env` as an argument.

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -29,7 +30,8 @@ function BayesianNetwork{V}() where {V}
...
+        (;),    # Empty NamedTuple for loop_vars
```
sunxd3, line 34: repeating the comment before, this should probably be a dictionary

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -54,7 +56,7 @@ function translate_BUGSGraph_to_BayesianNetwork(g::JuliaBUGS.BUGSGraph, evaluati
     # Preallocate arrays/dictionaries.
     names = Vector{VarName}(undef, n)
     names_to_ids = Dict{VarName,Int}()
-    values = Dict{VarName,Any}()
+    loop_vars = NamedTuple()
```
sunxd3, line 59: this too, prob should be a dictionary

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -66,10 +68,10 @@ function translate_BUGSGraph_to_BayesianNetwork(g::JuliaBUGS.BUGSGraph, evaluati
     for (i, varname) in enumerate(varnames)
         nodeinfo = g[varname]
         names[i] = varname
-        values[varname] = AbstractPPL.get(evaluation_env, varname)
         names_to_ids[varname] = i
         is_stochastic[i] = nodeinfo.is_stochastic
         is_observed[i] = nodeinfo.is_observed
+        loop_vars = merge(loop_vars, nodeinfo.loop_vars)
```
sunxd3, line 74: `loop_vars` should be a var-individual property.

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -161,3 +164,39 @@ function add_edge!(bn::BayesianNetwork{V,T}, from::V, to::V)::Bool where {T,V}
...
+            value = getproperty(evaluation_env, Symbol(varname))
+            bijector = Bijectors.bijector(dist_fn)
+            value_transformed = Bijectors.transform(bijector, value)
+
+            logpdf_val = Distributions.logpdf(dist_fn, value)
+            logjac = Bijectors.logabsdetjac(Bijectors.inverse(bijector), value_transformed)
+            logp += logpdf_val + logjac
+
+        else
+            fn = bn.deterministic_functions[i](evaluation_env, bn.loop_vars)
```
sunxd3, line 186: this would need fix, too

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -161,3 +164,39 @@ function add_edge!(bn::BayesianNetwork{V,T}, from::V, to::V)::Bool where {T,V}
...
 end
+
+function evaluate(bn::BayesianNetwork)
+    logp = 0.0
+    evaluation_env = bn.evaluation_env
+
+    for (i, varname) in enumerate(bn.names)
+        is_stochastic = bn.is_stochastic[i]
+        if is_stochastic
+            dist_fn = bn.distributions[i](evaluation_env, bn.loop_vars)
```
sunxd3, line 175: again, `loop_vars` should be different for each `varname`

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -161,3 +164,39 @@ function add_edge!(bn::BayesianNetwork{V,T}, from::V, to::V)::Bool where {T,V}
...
+
+            value = getproperty(evaluation_env, Symbol(varname))
```
sunxd3, line 177: `AbstractPPL.get` should work

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -161,3 +164,39 @@ function add_edge!(bn::BayesianNetwork{V,T}, from::V, to::V)::Bool where {T,V}
...
+            bijector = Bijectors.bijector(dist_fn)
+            value_transformed = Bijectors.transform(bijector, value)
+
+            logpdf_val = Distributions.logpdf(dist_fn, value)
+            logjac = Bijectors.logabsdetjac(Bijectors.inverse(bijector), value_transformed)
+            logp += logpdf_val + logjac
+
+        else
+            fn = bn.deterministic_functions[i](evaluation_env, bn.loop_vars)
+            parent_vals = parent_values(bn, evaluation_env, i)
```
sunxd3, line 187: this is no longer needed, because now we don't store values for each variables, but managing all the values in `evaluation_env`

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -161,3 +164,39 @@ function add_edge!(bn::BayesianNetwork{V,T}, from::V, to::V)::Bool where {T,V}
...
+
+            # Update NamedTuple in an immutable way
+            evaluation_env = merge(evaluation_env, NamedTuple{(Symbol(varname),)}((fn,)))
```
sunxd3, line 190: look at https://github.com/TuringLang/JuliaBUGS.jl/blob/4f8da8424b75dafee7dd237f022b3acd6ae6369b/src/model.jl#L587

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -161,3 +164,39 @@ function add_edge!(bn::BayesianNetwork{V,T}, from::V, to::V)::Bool where {T,V}
...
+        end
+    end
+    return evaluation_env, logp
+end
+
+function parent_values(bn::BayesianNetwork, evaluation_env::NamedTuple, i::Int)
```
sunxd3, line 196: this is probably not needed anymore

src/experimental/ProbabilisticGraphicalModels/conditioning.jl
```diff
@@ -7,8 +7,10 @@ function condition(
     bn::BayesianNetwork{V}, conditioning_variables_and_values::Dict{V,<:Any}
 ) where {V}
     is_observed = copy(bn.is_observed)
-    values = copy(bn.values)
-    bn_new = BangBang.setproperties!!(bn; is_observed=is_observed, values=values)
+    evaluation_env = merge(bn.evaluation_env, NamedTuple(conditioning_variables_and_values))  # Merge into a NamedTuple
```
sunxd3, line 10: `namedTuple` will error when the Dictionary's key type is `VarName`

src/experimental/ProbabilisticGraphicalModels/conditioning.jl
```diff
@@ -29,29 +33,57 @@ function condition!(
...
+"""
+    decondition(bn::BayesianNetwork{V}, variables::Vector{V}) where {V}
+
+Remove conditioning from a subset of variables in the Bayesian Network.
+"""
 function decondition(bn::BayesianNetwork{V}, variables::Vector{V}) where {V}
     is_observed = copy(bn.is_observed)
-    values = copy(bn.values)
-    bn_new = BangBang.setproperties!!(bn; is_observed=is_observed, values=values)
+    evaluation_env = NamedTuple(
```
sunxd3, line 68: `evaluation_env` contains **all** the values always, no need to remove variables during unconditioning

test/experimental/ProbabilisticGraphicalModels/bayesnet.jl
```diff
@@ -368,4 +371,64 @@ using JuliaBUGS: @bugs, compile, NodeInfo, VarName
...
+
+        @test bn.is_stochastic[bn.names_to_ids[VarName(:a)]] == true
+        @test bn.is_stochastic[bn.names_to_ids[VarName(:b)]] == true
+        @test bn.is_stochastic[bn.names_to_ids[VarName(:c)]] == true
+
+        @test bn.distributions[bn.names_to_ids[VarName(:a)]] isa Function
+        @test bn.distributions[bn.names_to_ids[VarName(:b)]] isa Function
+        @test bn.distributions[bn.names_to_ids[VarName(:c)]] isa Function
+    end
+    @testset "Translating Complex BUGSGraph to BayesianNetwork and Evaluate" begin
```
sunxd3, line 405: let's add one example with for loops, otherwise many things are not tested

sunxd3 (changes requested):

src/experimental/ProbabilisticGraphicalModels/conditioning.jl
```diff
@@ -7,8 +7,16 @@ function condition(
     bn::BayesianNetwork{V}, conditioning_variables_and_values::Dict{V,<:Any}
 ) where {V}
     is_observed = copy(bn.is_observed)
-    values = copy(bn.values)
-    bn_new = BangBang.setproperties!!(bn; is_observed=is_observed, values=values)
+    evaluation_env = merge(
+        bn.evaluation_env,
+        NamedTuple{Tuple(Symbol(k) for k in keys(conditioning_variables_and_values))}(
```
sunxd3, line 12: key of `conditioning_variables_and_values` can be VarName, here just use AbstractPPL.set should work

src/experimental/ProbabilisticGraphicalModels/conditioning.jl
```diff
@@ -29,28 +39,50 @@ function condition!(
         else
             bn.is_observed[id] = true
         end
-        bn.values[name] = value
+        new_evaluation_env = merge(new_evaluation_env, (name => value,))
```
sunxd3, line 42: same issue here

test/experimental/ProbabilisticGraphicalModels/bayesnet.jl
```diff
@@ -368,4 +369,88 @@ using JuliaBUGS: @bugs, compile, NodeInfo, VarName
...
+        evaluation_env, logp = evaluate(complex_bn)
+
+        @test haskey(evaluation_env, :a)
+        @test haskey(evaluation_env, :b)
+        @test haskey(evaluation_env, :c)
+        @test evaluation_env[:c] ≈ evaluation_env[:a] + evaluation_env[:b]
+        @test logp ≈
+            logpdf(Normal(0, 1), evaluation_env[:a]) +
+              logpdf(Normal(1, 1), evaluation_env[:b])
+    end
```
sunxd3, line 431:
```suggestion
    end
   
```

sunxd3: extra line for clarity

sunxd3 (changes requested):
tiny things, we're really close

sunxd3 (approved):
@naseweisssss thanks a lot for this! Looks good to me now.

PR #278, opened 2025-02-27
Generating Julia function for log density evaluation

sunxd3:
description to come...

### DIFF
--- a/src/source_gen.jl
+++ b/src/source_gen.jl
@@ -1,0 +1,380 @@
+function _build_stmt_to_stmt_id(
+    block_expr, stmt_to_stmt_id=IdDict{Expr,Int}(), next_id=Ref(1)
+)
+    @assert Meta.isexpr(block_expr, :block)
+    for statement in block_expr.args
+        if Meta.isexpr(statement, (:(=), :call))
+            stmt_to_stmt_id[statement] = next_id[]
+            next_id[] += 1
+        elseif Meta.isexpr(statement, :for)
+            stmt_to_stmt_id = _build_stmt_to_stmt_id(
+                statement.args[2], stmt_to_stmt_id, next_id
+            )
+        end
+    end
+    return stmt_to_stmt_id
+end
+
+# like deepcopy, but stop at :(=) and :call level (:call in :for don't matter)
+function _copy_model_def(model_def, new_model_def=Expr(:block))
+    @assert Meta.isexpr(model_def, :block)
+    for statement in model_def.args
+        if Meta.isexpr(statement, (:(=), :call))
+            push!(new_model_def.args, statement)
+        elseif Meta.isexpr(statement, :for)
+            # Recursively copy the inner block of the for-loop.
+            new_body = _copy_model_def(statement.args[2], Expr(:block))
+            new_for = Expr(:for, statement.args[1], new_body)
+            push!(new_model_def.args, new_for)
+        else
+            # For any other kind of expression, simply include it as is.
+            push!(new_model_def.args, statement)
+        end
+    end
+    return new_model_def
+end
+
+function _build_stmt_id_to_stmt(stmt_to_stmt_id::IdDict{Expr,Int})
+    stmt_id_to_stmt = IdDict{Int,Expr}()
+    for (stmt, stmt_id) in stmt_to_stmt_id
+        stmt_id_to_stmt[stmt_id] = stmt
+    end
+    return stmt_id_to_stmt
+end
+
+struct StatementIdAttributePass{ET} <: CompilerPass
+    all_variables_in_graph::Set{<:VarName}
+    stmt_ids::IdDict{Expr,Int}
+    env::ET
+
+    var_to_stmt_id::Dict{VarName,Int}
+end
+
+function analyze_statement(
+    pass::StatementIdAttributePass, expr::Expr, loop_variables::NamedTuple
+)
+    lhs_expression = is_deterministic(expr) ? expr.args[1] : expr.args[2]
+    merged_env = merge(pass.env, loop_variables)
+    simplified_lhs = simplify_lhs(merged_env, lhs_expression)
+    varname = if simplified_lhs isa Symbol
+        VarName{simplified_lhs}()
+    else
+        symbol, indices... = simplified_lhs
+        VarName{symbol}(IndexLens(indices))
+    end
+
+    if varname in pass.all_variables_in_graph
+        if haskey(pass.var_to_stmt_id, varname)
+            if JuliaBUGS.is_stochastic(expr)
+                # could be transformed data, so the variable should be associated with the stochastic statement
+                pass.var_to_stmt_id[varname] = pass.stmt_ids[expr]
+            end
+        else
+            pass.var_to_stmt_id[varname] = pass.stmt_ids[expr]
+        end
+    end
+    return nothing
+end
+
+function _build_var_to_stmt_id(model::BUGSModel, stmt_ids::IdDict{Expr,Int})
+    pass = StatementIdAttributePass(
+        Set(labels(model.g)), stmt_ids, model.evaluation_env, Dict{VarName,Int}()
+    )
+    analyze_block(pass, model.model_def)
+    return pass.var_to_stmt_id
+end
+
+# coarse graph may contains nodes whose degree is 0, these are transformed data
+function _build_coarse_dep_graph(
+    model::BUGSModel, stmt_to_stmt_id::IdDict{Expr,Int}, var_to_stmt_id::Dict{VarName,Int}
+)
+    fine_graph = model.g
+    coarse_graph = Graphs.SimpleDiGraph(length(stmt_to_stmt_id))
+    for edge in Graphs.edges(fine_graph.graph)
+        src_varname = label_for(fine_graph, src(edge))
+        dst_varname = label_for(fine_graph, dst(edge))
+        src_stmt_id = var_to_stmt_id[src_varname]
+        dst_stmt_id = var_to_stmt_id[dst_varname]
+        add_edge!(coarse_graph, src_stmt_id, dst_stmt_id)
+    end
+    return coarse_graph
+end
+
+function _copy_and_remove_stmt_with_degree_0(
+    model_def, stmt_to_stmt_id, coarse_graph, new_model_def=Expr(:block)
+)
+    @assert Meta.isexpr(model_def, :block)
+    for statement in model_def.args
+        if Meta.isexpr(statement, (:(=), :call))
+            if degree(coarse_graph, stmt_to_stmt_id[statement]) != 0
+                push!(new_model_def.args, statement)
+            end
+        elseif Meta.isexpr(statement, :for)
+            # Recursively copy the inner block of the for-loop.
+            new_body = _copy_and_remove_stmt_with_degree_0(
+                statement.args[2], stmt_to_stmt_id, coarse_graph, Expr(:block)
+            )
+            new_for = Expr(:for, statement.args[1], new_body)
+            push!(new_model_def.args, new_for)
+        else
+            # For any other kind of expression, simply include it as is.
+            push!(new_model_def.args, statement)
+        end
+    end
+    return new_model_def
+end
+
+# represent loop variables as (loop_var, lower_bound, upper_bound)
+function _fully_fission_loop(
+    expr, stmt_to_stmt_id, evaluation_env, current_loop=(), fissioned_stmts=[]
+)
+    for stmt in expr.args
+        if Meta.isexpr(stmt, (:(=), :call)) # :call is for ~
+            push!(fissioned_stmts, (current_loop, [stmt]))
+        elseif Meta.isexpr(stmt, :for)
+            MacroTools.@capture(stmt.args[1], loop_var_ = l_:h_)
+            _fully_fission_loop(
+                stmt.args[2],
+                stmt_to_stmt_id,
+                evaluation_env,
+                (current_loop..., (loop_var, l, h)),
+                fissioned_stmts,
+            )
+        end
+    end
+    return fissioned_stmts
+end
+
+function _sort_fissioned_stmts(stmt_dep_graph, fissioned_stmts, stmt_ids)
+    sorted_stmts = Graphs.topological_sort(stmt_dep_graph)
+    sorted_fissioned_stmts = []
+    for stmt_id in sorted_stmts
+        for (loops, stmt) in fissioned_stmts
+            if stmt_ids[first(stmt)] == stmt_id
+                push!(sorted_fissioned_stmts, (loops, stmt))
+            end
+        end
+    end
+    return sorted_fissioned_stmts
+end
+
+function _reconstruct_model_def_from_sorted_fissioned_stmts(sorted_fissioned_stmts)
+    args = []
+    for (loops, stmt) in sorted_fissioned_stmts
+        if loops == ()
+            push!(args, first(stmt))
+        else
+            push!(args, __gen_loop_expr(loops, first(stmt)))
+        end
+    end
+    return Expr(:block, args...)
+end
+
+function __gen_loop_expr(loop_vars, stmt)
+    loop_var, l, h = loop_vars[1]
+    if length(loop_vars) == 1
+        return MacroTools.@q for $(loop_var) in ($(l)):($(h))
+            $(stmt)
+        end
+    else
+        return MacroTools.@q for $(loop_var) in ($(l)):($(h))
+            $(__gen_loop_expr(loop_vars[2:end], stmt))
+        end
+    end
+end
+
+function can_reorder(coarse_graph::Graphs.SimpleDiGraph)
+    if Graphs.is_cyclic(coarse_graph)
+        return false
+    end
+    return true
+end
+
+function _lower_model_def_to_represent_observe_stmts(
+    reconstructed_model_def,
+    stmt_to_stmt_id,
+    stmt_types,
+    evaluation_env,
+    lowered_model_def=Expr(:block),
+)
+    for statement in reconstructed_model_def.args
+        if Meta.isexpr(statement, (:(=), :call))
+            if stmt_types[stmt_to_stmt_id[statement]] == :observed
+                MacroTools.@capture(statement, lhs_ ~ rhs_)
+                new_stmt = MacroTools.@q $(lhs) ≂ $(rhs)
+                push!(lowered_model_def.args, new_stmt)
+            else
+                push!(lowered_model_def.args, statement)
+            end
+        elseif Meta.isexpr(statement, :for)
+            # Recursively copy the inner block of the for-loop.
+            new_body = _lower_model_def_to_represent_observe_stmts(
+                statement.args[2], stmt_to_stmt_id, stmt_types, evaluation_env, Expr(:block)
+            )
+            new_for = Expr(:for, statement.args[1], new_body)
+            push!(lowered_model_def.args, new_for)
+        else
+            # For any other kind of expression, simply include it as is.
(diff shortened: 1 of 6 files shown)


sunxd3:
@mhauru @yebai 

Realistically, I don't think anyone can give this a proper review. But could you spend a little bit of time just to check for some glaring issues?

The major changes are in the file [src/source_gen.jl](https://github.com/TuringLang/JuliaBUGS.jl/pull/278/files#diff-1bf54a5ffb7d8c8cb17a4ea6fb69d7a97ffdd21b1a1eccd9b45254649ae4fdb1).

If I got some time that I don't have, I'll break it down into multiple PRs. Really sorry for not be able to do this.

sunxd3:
The failures are on v1.12

yebai:
The benchmark numbers for Mooncake look weird. Please take a look.

sunxd3:
```julia
f = Base.Fix1(model.log_density_computation_function, model.evaluation_env)
p = Base.Fix1(LogDensityProblems.logdensity, model)
```

where 

```julia
function _eval_logdensity(model, ::UseGeneratedLogDensityFunction, x)
    return model.log_density_computation_function(model.evaluation_env, x)
end

# not used in this case
function _eval_logdensity(model, ::UseGraph, x)
    _, logp = JuliaBUGS.evaluate!!(model, x)
    return logp
end

function LogDensityProblems.logdensity(model::BUGSModel, x::AbstractArray)
    return _eval_logdensity(model, model.evaluation_mode, x)
end
```

benchmark on `f`

```
BenchmarkTools.Trial: 10000 samples with 6 evaluations per sample.
 Range (min … max):  5.048 μs …  10.771 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     5.826 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   5.817 μs ± 329.168 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                      ▄██▇▁                                    
  ▁▁▂▂▃▃▃▃▂▂▂▂▂▂▂▂▂▂▃▆█████▇▅▄▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  5.05 μs         Histogram: frequency by time        7.13 μs <

 Memory estimate: 1.19 KiB, allocs estimate: 6.
```

benchmark on `p`
```julia
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  119.416 μs …  12.799 ms  ┊ GC (min … max): 0.00% … 98.59%
 Time  (median):     139.667 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   148.119 μs ± 215.937 μs  ┊ GC (mean ± σ):  5.26% ±  4.13%

                  ▂▁▃▄▄▅▇█▇▆▅▄▁                                  
  ▁▁▂▃▄▅▅▅▆▆▆▆▇████████████████▇▆▄▃▃▃▃▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  119 μs           Histogram: frequency by time          176 μs <

 Memory estimate: 180.44 KiB, allocs estimate: 4791.
```



I had a discussion with @willtebbutt last week, the benchmark seems to be more of a Mooncake issue than JuliaBUGS. We are not one hundred percent, and will try to solve this soon.

yebai:
I don't understand why it is related to Mooncake. The above example doesn't use `Mooncake`.

sunxd3:
benchmark code for gradient is added

yebai:
@penelopeysm @mhauru, can you review this PR to the best of your ability? I'll take a look more carefully later.

sunxd3:
Thanks @penelopeysm , just to reiterate, I know the amount of lines changed is very large. But only https://github.com/TuringLang/JuliaBUGS.jl/blob/sunxd/source_gen/src/source_gen.jl and its test https://github.com/TuringLang/JuliaBUGS.jl/blob/sunxd/source_gen/test/source_gen.jl are worth looking. Others are either low risk or updates to benchmarking that's no super relevant.

penelopeysm:
I didn't look at any of the Julia files because I don't know how to, and I have no comments on the other files!

Just kidding. I'll do my best haha

docs/src/source_gen.md
```diff
@@ -0,0 +1,481 @@
...
+The computation of the dependence vectors is done in the following steps:
+When executing the for loop from `i = 2` to `i = T`, we need to compute the dependence vector. Let's examine the case when `i = 2`:
+
+- LHS: `s[2]` is defined at iteration `i=2`
+- RHS: `s[1]` is defined at iteration `i=1`
+- Dependence vector: $\vec{d} = \vec{j}_{sink} - \vec{i}_{source} = (2) - (1) = (1)$
+
+All subsequent iterations follow the same pattern and have the same dependence vector of $(1)$. Because all dependence vectors are lexicographically non-negative, the loop is sequentially valid.
+
+This requires storing the loop variable `i` for each variable, but we already computed this with JuliaBUGS compilation, so not much overhead is required.
```
penelopeysm, line 278: To make sure I'm understanding this properly: as part of the compilation, you step through the loop and on each iteration you calculate the dependency vector for the variables in the loop?

docs/src/source_gen.md
```diff
@@ -0,0 +1,481 @@
...
+
+If the statement dependence graph is acyclic, then we topologically sort the statements and fission all the loops to create a sequentially valid program.
+
+In case the statement dependence graph is not acyclic.
+If the statements that form a cycle are all from the same loop (potentially at different nested levels), then we compute the dependence vectors to determine if the loop is sequentially valid. 
+
+Otherwise, the program need to be rewritten.
+
+We don't attempt to apply further transformations to the program, because it is a hard problem. And to give a flavor of the difficulty of rewriting programs, let's consider the following example:
+
```
penelopeysm, line 298: If I understand right, the examples that follow this sentence are programmes that will be rejected?

If so, I think it might be worth making this really explicit.

docs/src/source_gen.md
```diff
@@ -0,0 +1,481 @@
...
+    S3_3 --> S2
+    
+    subgraph "Loop dependencies"
+        S3_1 --> S3_2
+        S3_2 --> S3_3
+        S3_3 --> S3_1
+    end
+```
+
+This represents a worst case where we can't do much better than fully unrolling.
```
penelopeysm, line 416: (Continued from above) Because this sentence doesn't entirely make it clear whether the model can be run after the loop is unrolled. As I understand it, this dependency graph is cyclic (even after unrolling).

docs/src/source_gen.md
```diff
@@ -0,0 +1,481 @@
...
+    end
+    for i = 1:30
+        for j = 1:5
+            Y[i, j] \eqsim dnorm(mu[i, j], var"tau.c")
+        end
+    end
+end
+```
+
+We made a simple change to the program to prepare for lowering: we need to distinguish between observations and model parameters (because they correspond to different code). We introduce a new operator into the program `\eqsim` to indicate that the left hand side is an observation.
```
penelopeysm, line 450: Personally, I don't think I'm entirely comfortable with eqsim; I think it looks too much like an ordinary equals, and also it's hard to type.

I think something like `~=` would be better for my eyes - would you be ok with that?

penelopeysm (commented):
I haven't yet looked at the code, but I've given the documentation a read through. It's very good! I just put in a couple of comments on places where I still had a bit of confusion. I'll look at the source code another time.

docs/src/source_gen.md
```diff
@@ -0,0 +1,481 @@
...
+        y[i] ~ normal(x[2], i)  # (2) 
+    end
+end
+```
+
+Each execution of the loop body corresponds to an *iteration vector* $\vec{k} = (i, j)$. The set of all possible iteration vectors is the *iteration space*, here $\{(i, j) | 1 \le i \le 2, 1 \le j \le 3\}$. Iterations execute sequentially in lexicographical order: (1,1), (1,2), (1,3), (2,1), (2,2), (2,3).
+
+**Dependence Vectors**
+
+If a statement execution at iteration $\vec{i}$ (source) defines a value used by a statement execution at iteration $\vec{j}$ (sink), the *dependence vector* is $\vec{d} = \vec{j} - \vec{i}$. It represents the distance between dependent iterations. (Sometimes, we don't care about the number, so we can simply use the sign of the dependence vector.)
```
sunxd3, line 175:
```suggestion
If a statement execution at iteration $\vec{i}$ (source) defines a value used by a statement execution at iteration $\vec{j}$ (sink), the *dependence vector* is $\vec{d} = \vec{j} - \vec{i}$. It represents the distance between dependent iterations. (Sometimes, we don't care about the distance between define and use, so we can simply use the sign of the elements of the dependence vector to represent the dependence relation.)
```

docs/src/source_gen.md
sunxd3: Currently, we don't compute the dependence vector. But we store the values of loop inductions variables for each variables in the model.

docs/src/source_gen.md
sunxd3: That's correct. Thanks!

docs/src/source_gen.md
```diff
@@ -0,0 +1,481 @@
...
+First a statement dependence graph is computed.
+
+If the statement dependence graph is acyclic, then we topologically sort the statements and fission all the loops to create a sequentially valid program.
+
+In case the statement dependence graph is not acyclic.
+If the statements that form a cycle are all from the same loop (potentially at different nested levels), then we compute the dependence vectors to determine if the loop is sequentially valid. 
+
+Otherwise, the program need to be rewritten.
+
+We don't attempt to apply further transformations to the program, because it is a hard problem. And to give a flavor of the difficulty of rewriting programs, let's consider the following example:
```
sunxd3, line 297:
```suggestion
We don't attempt to apply further transformations to the program, because it is a hard problem. We will use the following example to show why program transformations can be a difficult task. We will not attempt to implement the transformation demonstrated here.

Consider this model,
```

docs/src/source_gen.md
sunxd3: I updated the example to make it more clear (the fully unrolled version should not be cyclic, unless I made a dumb mistake)

src/source_gen.jl
```diff
@@ -0,0 +1,602 @@
...
+        return result
+    end
+end
+
+function _generate_lowered_model_def(model_def, g::JuliaBUGS.BUGSGraph, evaluation_env)
+    stmt_to_stmt_id = _build_stmt_to_stmt_id(model_def)
+    stmt_id_to_stmt = _build_stmt_id_to_stmt(stmt_to_stmt_id)
+    var_to_stmt_id = _build_var_to_stmt_id(model_def, g, evaluation_env, stmt_to_stmt_id)
+    stmt_id_to_var = _build_stmt_id_to_var(var_to_stmt_id)
+    coarse_graph = _build_coarse_dep_graph(g, stmt_to_stmt_id, var_to_stmt_id)
```
penelopeysm, line 313: I like the flow of this function a lot! Nothing to nitpick on. Functions are clearly named too.

It was fairly easy for me to insert `@show` statements everywhere to see what was going on. However, in the docs that you have, I wonder if it might be useful to also demonstrate (with code) what's going on in each of these steps? I don't think it's super urgent, as I said it wasn't hard to play around with it myself. But it'd be a nice way of walking a reader through the transformations.

penelopeysm (changes requested):
The only thing I really think is important is the macro hygiene one!

src/source_gen.jl
```diff
@@ -0,0 +1,602 @@
...
+
+function __gen_deterministic_exprs(stmt)
+    return stmt
+end
+
+function __gen_model_parameter_exprs(stmt)
+    MacroTools.@capture(stmt, lhs_ ~ rhs_)
+    return MacroTools.@q begin
+        __dist__ = $rhs
+        __b__ = Bijectors.bijector(__dist__)
```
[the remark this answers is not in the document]
penelopeysm: (Link to relevant section in DynamicPPL: https://github.com/TuringLang/DynamicPPL.jl/blob/019e41b341ccfb04b8467e50bee6918fbe1d74c0/src/compiler.jl#L436)

src/model.jl
```diff
@@ -42,14 +42,25 @@ function FlattenedGraphNodeData(
     )
 end
 
+abstract type EvaluationMode end
+
+struct UseGeneratedLogDensityFunction <: EvaluationMode end
+struct UseGraph <: EvaluationMode end
```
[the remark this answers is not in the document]
sunxd3: No particular reason, just didn't come to my mind at the time.

I don't think `enum` is more elegant than subtyping for this case (not saying it's worse either). Do you have a preference?

src/model.jl
```diff
@@ -195,6 +209,39 @@ function BUGSModel(
...
+        log_density_computation_function = eval(log_density_computation_expr)
+        pass = CollectSortedNodes(evaluation_env)
+        JuliaBUGS.analyze_block(pass, reconstructed_model_def)
+        sorted_nodes = pass.sorted_nodes
+        original_parameters_length = length(parameters)
+        parameters = VarName[vn for vn in sorted_nodes if vn in parameters]
+        @assert length(parameters) == original_parameters_length "there are less parameters in the generated log density function than in the original model"
+        flattened_graph_node_data = FlattenedGraphNodeData(g, sorted_nodes)
+    else
+        log_density_computation_function = identity
```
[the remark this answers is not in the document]
sunxd3: yeah, that's fair

src/model.jl
```diff
@@ -378,6 +429,16 @@ function settrans(model::BUGSModel, bool::Bool=!(model.transformed))
     return BangBang.setproperty!!(model, :transformed, bool)
 end
 
+function set_evaluation_mode(model::BUGSModel, mode::EvaluationMode)
```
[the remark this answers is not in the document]
sunxd3: In this case, because it never modify the original object, I didn't use any `!`.
In my mind, `!!` means "mutate if possible". Then if no way of mutating anyway, maybe okay to not use `!` at all?

test/source_gen.jl
```diff
@@ -0,0 +1,91 @@
...
+    params = JuliaBUGS.getparams(model_with_consistent_sorted_nodes)
+    result_with_bugsmodel = JuliaBUGS.evaluate!!(
+        model_with_consistent_sorted_nodes, params
+    )[2]
+    result_with_log_density_computation_function = log_density_computation_functions[example_name](
+        evaluation_envs[example_name], params
+    )
+    @test result_with_old_model ≈ result_with_bugsmodel
+    @test result_with_log_density_computation_function ≈ result_with_bugsmodel
+end
```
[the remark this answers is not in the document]
sunxd3: I don't think so unfortunately. Because I am generating function and eval them. If combined, there are world-age problems.

src/BUGSExamples/Volume_3/03_Fire.jl
```diff
@@ -27,7 +27,7 @@ model_def = @bugs begin
 
     # xf prediction from fitted distribution
     xf = xa * delta + xb * (1 - delta)
-    xa ~ dlnorm(mu, tau)[:, theta]
+    xa ~ truncated(dlnorm(mu, tau), nothing, theta)
```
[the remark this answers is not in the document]
sunxd3: that's totally fine, thanks a lot for taking a look already! The examples are relatively low-risk, the mistakes can be fixed later when you emerge.

src/source_gen.jl
sunxd3: agree that would be very helpful, I'll do this later

src/source_gen.jl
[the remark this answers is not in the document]
sunxd3: this is done now

sunxd3:
FYI, I switched to directly using the generated function for benchmarking instead of through `LogDensityProblems`

src/model.jl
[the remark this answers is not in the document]
penelopeysm: No not really, just curious because I think I saw you used it elsewhere in the code! I guess the type approach is generally better for type stability since you can dispatch on it, rather than enums which are just values belonging to the same type (well I guess you can use `Val{...}` but... really who wants to do that)

src/model.jl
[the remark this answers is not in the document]
penelopeysm: Ah I see what you mean. I guess I didn't look closely at the definition of a `Model` and I just saw `setproperty!!`, so it seemed to me that this should 'inherit' it. I think what you say makes sense though so happy to keep it this way 👍

test/source_gen.jl
```diff
@@ -0,0 +1,99 @@
...
```
penelopeysm, line 91:
```suggestion
@testset "source_gen: $example_name" for example_name in test_examples
    model, evaluation_env = _create_model(example_name)
    lowered_model_def, reconstructed_model_def = _generate_lowered_model_def(
        model.model_def, model.g, evaluation_env
    )
    log_density_computation_expr = _gen_log_density_computation_function_expr(
        lowered_model_def, evaluation_env, gensym(example_name)
    )
    log_density_computation_function = eval(log_density_computation_expr)

    model_with_consistent_sorted_nodes = _create_bugsmodel_with_consistent_sorted_nodes(
        model, reconstructed_model_def
    )
    result_with_old_model = JuliaBUGS.evaluate!!(model)[2]
    params = JuliaBUGS.getparams(model_with_consistent_sorted_nodes)
    result_with_bugsmodel = JuliaBUGS.evaluate!!(
        model_with_consistent_sorted_nodes, params
    )[2]
    result_with_log_density_computation_function = log_density_computation_function(
        evaluation_env, params
    )
    @test result_with_old_model ≈ result_with_bugsmodel
    @test result_with_log_density_computation_function ≈ result_with_bugsmodel
end
```

I might be missing something as i don't have a very deep understanding of Julia's world age, but this seems to pass locally for me (tested on both 1.10 and 1.11).

src/source_gen.jl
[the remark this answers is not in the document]
penelopeysm: Ahhh, okay, I see what you mean with the other functions too, there's flattened_values and stuff like that.

I guess I'd say that this is something that in an ideal world should be fixed, but not opposed to merging in its current state especially with the check on the reserved names 🙂

penelopeysm:
Just the comment about the testset, then happy to click approve!

docs/src/source_gen.md
sunxd3: absolutely agree -- I think Unicode characters can be good to look at, but annoying to type (particularly if not in an IDE)

right now, this is pretty internal, but maybe JuliaBUGS can expose this in the future, then this "lowering" function can just turn a Julia AST into a logdensity function. but we can have a proper discussion on what to use thn

src/model.jl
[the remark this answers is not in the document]
sunxd3: My enum use in Julia is still impacted by C. 

I might do `enum Color; Red; Green; ...` just this case it doesn't feel as intuitive as directly subtyping for me.

sunxd3:
Thanks Penny, regarding the testset, it indeed passes.
I thought it might have world-age issue because that's the case in the benchmark code. Maybe `testset` is special in some sense? I am not sure.
Changed to your suggestions on this, really appreciate it!

penelopeysm (approved):

yebai:
I suggest we include the primal time for graph-based log density evaluation in the results table for JuliaBUGS Mooncake, i.e: 

- Density Time (Graph approach; µs)
- Density Time (Code generation approach; µs)
- Density+Gradient Time (µs) -- We only need gradient time for the code generation approach for brevity.

yebai:
Excellent work, @sunxd3 and @penelopeysm!

PR #289, opened 2025-03-30
Evaluate Function with Recursive Approach to Discreteness

naseweisssss:
- Marginalization implementation (recursion-based) for discrete variables in Bayesian networks
- Test case included
- There should be another PR with similar implementation but with DP.


test/experimental/ProbabilisticGraphicalModels/bayesnet.jl
```diff
@@ -638,4 +639,643 @@ using AbstractPPL
...
+                    bn.deterministic_functions,
+                    bn.stochastic_ids,
+                    bn.deterministic_ids,
+                    bn.is_stochastic,
+                    bn.is_observed,
+                    new_node_types,
+                    bn.transformed_var_lengths,
+                    bn.transformed_param_length,
+                )
+            end
```
sunxd3, line 1216: duplicated function?

sunxd3:
@naseweisssss this looks really good. 

Two questions:
* What is the purpose of returning `final_param_idx`?
* was https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L368-L373 preventitive, or fix because something went wrong so you have to add this spacial handling?

naseweisssss:
> @naseweisssss this looks really good.
> 
> Two questions:
> 
> * What is the purpose of returning `final_param_idx`?
> * was https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L368-L373
>    preventitive, or fix because something went wrong so you have to add this spacial handling?

The `final_param_idx` is mainly used when processing a continuous unobserved node.

https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L376-L383

It is used to keep track of the parameter array index we are currently on now. However, the `parameter_values` vector needs to have parameters in the exact same order that they'll be consumed by the algorithm

naseweisssss:
> @naseweisssss this looks really good.
> 
> Two questions:
> 
> * What is the purpose of returning `final_param_idx`?
> * was https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L368-L373
>    preventitive, or fix because something went wrong so you have to add this special handling?


For line 368 - 373, I am not sure if I still need it since we assume all variables is in transformed space. It was meant to:

Slice the right number of values from the parameter vector
Transform them back to the original space
Evaluate the model's log probability

sunxd3:
> > * What is the purpose of returning `final_param_idx`?
> 
> The `final_param_idx` is mainly used when processing a continuous unobserved node.
> 
> https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L376-L383
> 
> It is used to keep track of the parameter array index we are currently on now. However, the `parameter_values` vector needs to have parameters in the exact same order that they'll be consumed by the algorithm

My confusion is what returning `final_param_idx` provides beyond `param_idx` and `var_length`. Maybe my intuition is wrong, but if we assume the order parameter values are consumed (which we do), and also the length of `parameter_values` doesn't change, then it's not necessary to return the final param_idx, is this wrong?

sunxd3:
> > * was https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L368-L373
> >   
> >   preventitive, or fix because something went wrong so you have to add this spacial handling?
> 
> Its supposed to:
> 
> Slice the right number of values from the parameter vector Transform them back to the original space Evaluate the model's log probability
> 
> This is used when a parameter vector is given.

This I get, my original comment is that the reason we fetch individual transformed length of variables is that we assume they never change, and the total length doesn't change. `var_lengths` already contains all the transformed lengths.

This is why I am asking if there is an error during testing without adding these lines.

naseweisssss:
> > > * What is the purpose of returning `final_param_idx`?
> > 
> > 
> > The `final_param_idx` is mainly used when processing a continuous unobserved node.
> > https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L376-L383
> > 
> > It is used to keep track of the parameter array index we are currently on now. However, the `parameter_values` vector needs to have parameters in the exact same order that they'll be consumed by the algorithm
> 
> My confusion is what returning `final_param_idx` provides beyond `param_idx` and `var_length`. Maybe my intuition is wrong, but if we assume the order parameter values are consumed (which we do), and also the length of `parameter_values` doesn't change, then it's not necessary to return the final param_idx, is this wrong?

I am thinking how would the parent call know how many parameters its child calls consumed, especially in complex networks with nested branches even if we assume we know the length of `parameter_values` and the order. I am not sure if I am thinking this right. What do you think?

naseweisssss:
> > > * was https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L368-L373
> > >   
> > >   preventitive, or fix because something went wrong so you have to add this spacial handling?
> > 
> > 
> > Its supposed to:
> > Slice the right number of values from the parameter vector Transform them back to the original space Evaluate the model's log probability
> > This is used when a parameter vector is given.
> 
> This I get, my original comment is that the reason we fetch individual transformed length of variables is that we assume they never change, and the total length doesn't change. `var_lengths` already contains all the transformed lengths.
> 
> This is why I am asking if there is an error during testing without adding these lines.

Yes this particular test would fail if this is removed.

https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/test/experimental/ProbabilisticGraphicalModels/bayesnet.jl#L742-L785

During this process, the algorithm needs to know the transformed length of each continuous variable. If this information isn't available for a particular execution path (which can happen in branch-dependent models like this one), the algorithm would fail without the fallback logic.

sunxd3:
> > > > * What is the purpose of returning `final_param_idx`?
> > > 
> > > 
> > > The `final_param_idx` is mainly used when processing a continuous unobserved node.
> > > https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L376-L383
> > > 
> > > It is used to keep track of the parameter array index we are currently on now. However, the `parameter_values` vector needs to have parameters in the exact same order that they'll be consumed by the algorithm
> > 
> > 
> > My confusion is what returning `final_param_idx` provides beyond `param_idx` and `var_length`. Maybe my intuition is wrong, but if we assume the order parameter values are consumed (which we do), and also the length of `parameter_values` doesn't change, then it's not necessary to return the final param_idx, is this wrong?
> 
> I am thinking how would the parent call know how many parameters its child calls consumed, especially in complex networks with nested branches even if we assume we know the length of `parameter_values` and the order. I am not sure if I am thinking this right. What do you think?

because we are dealing with **fixed-dimensional** models, all parameter values will eventually be consumed, right?

sunxd3:
> > > > * was https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L368-L373
> > > >   
> > > >   preventitive, or fix because something went wrong so you have to add this spacial handling?
> > > 
> > > 
> > > Its supposed to:
> > > Slice the right number of values from the parameter vector Transform them back to the original space Evaluate the model's log probability
> > > This is used when a parameter vector is given.
> > 
> > 
> > This I get, my original comment is that the reason we fetch individual transformed length of variables is that we assume they never change, and the total length doesn't change. `var_lengths` already contains all the transformed lengths.
> > This is why I am asking if there is an error during testing without adding these lines.
> 
> Yes this particular test would fail if this is removed.
> 
> https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/test/experimental/ProbabilisticGraphicalModels/bayesnet.jl#L742-L785
> 
> During this process, the algorithm needs to know the transformed length of each continuous variable. If this information isn't available for a particular execution path (which can happen in branch-dependent models like this one), the algorithm would fail without the fallback logic.
> 
> Do you think this is an error from the code?

Probably I misunderstood, in what way this model is a branch-dependent? Wouldn't all cont. variables (which is `y`?) have the same length throughout. Why did this fail?

naseweisssss:
> > > > > * What is the purpose of returning `final_param_idx`?
> > > > 
> > > > 
> > > > The `final_param_idx` is mainly used when processing a continuous unobserved node.
> > > > https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L376-L383
> > > > 
> > > > It is used to keep track of the parameter array index we are currently on now. However, the `parameter_values` vector needs to have parameters in the exact same order that they'll be consumed by the algorithm
> > > 
> > > 
> > > My confusion is what returning `final_param_idx` provides beyond `param_idx` and `var_length`. Maybe my intuition is wrong, but if we assume the order parameter values are consumed (which we do), and also the length of `parameter_values` doesn't change, then it's not necessary to return the final param_idx, is this wrong?
> > 
> > 
> > I am thinking how would the parent call know how many parameters its child calls consumed, especially in complex networks with nested branches even if we assume we know the length of `parameter_values` and the order. I am not sure if I am thinking this right. What do you think?
> 
> because we are dealing with **fixed-dimensional** models, all parameter values will eventually be consumed, right?

If I am understanding this right, should I precompute the parameter mapping instead of tracking the indices through recursive calls? This might potentially be more efficient for fixed-dimensional models.

naseweisssss:
> > > > > * was https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl#L368-L373
> > > > >   
> > > > >   preventitive, or fix because something went wrong so you have to add this spacial handling?
> > > > 
> > > > 
> > > > Its supposed to:
> > > > Slice the right number of values from the parameter vector Transform them back to the original space Evaluate the model's log probability
> > > > This is used when a parameter vector is given.
> > > 
> > > 
> > > This I get, my original comment is that the reason we fetch individual transformed length of variables is that we assume they never change, and the total length doesn't change. `var_lengths` already contains all the transformed lengths.
> > > This is why I am asking if there is an error during testing without adding these lines.
> > 
> > 
> > Yes this particular test would fail if this is removed.
> > https://github.com/TuringLang/JuliaBUGS.jl/blob/c736f872e9b07131dd692769f72dcf9b8a572c58/test/experimental/ProbabilisticGraphicalModels/bayesnet.jl#L742-L785
> > 
> > During this process, the algorithm needs to know the transformed length of each continuous variable. If this information isn't available for a particular execution path (which can happen in branch-dependent models like this one), the algorithm would fail without the fallback logic.
> > Do you think this is an error from the code?
> 
> Probably I misunderstood, in what way this model is a branch-dependent? Wouldn't all cont. variables (which is `y`?) have the same length throughout. Why did this fail?

Sorry I think my last reply is confusing. In this case, the test fails because when the bn is created from translating the BUGSGraph, the transformed_var_lengths dictionary might not contain entries for all variables. Hence, the fallback is handling the case where this info is not available.

sunxd3:
```julia
julia> compiled_model = compile(model_def, NamedTuple()) 
┌ Warning: Variables that can be directly computed from data are considered 'transformed data' in BUGS.
│ These variables won't appear in the probabilistic graph.
│ In this model, 'mu_z0' is one such variable.
│ There may be others, but this warning is only shown once.
│ If you're not working with existing BUGS programs, consider pre-computing these variables in Julia for efficiency.
└ @ JuliaBUGS ~/.julia/packages/JuliaBUGS/iDTyw/src/compiler_pass.jl:516
BUGSModel (parameters are in transformed (unconstrained) space, with dimension 2):

  Model parameters:
    y
    z

  Variable sizes and types:
    mu_z0: type = Float64
    mu: type = Float64
    sigma_z1: type = Float64
    y: type = Float64
    sigma_z0: type = Float64
    mu_z1: type = Float64
    z: type = Bool
    sigma: type = Float64

julia> compiled_model.transformed_var_lengths
Dict{AbstractPPL.VarName, Int64} with 2 entries:
  y => 1
  z => 1
```

all the variable lengths should be there, otherwise there is a bug.

Just to be clear, both (returning `final_param_idx` and adding to `var_lengths`) are necessary if we are dealing with a more general programming language. But because JuliaBUGS doesn't have control flow and all the variables should have the same length after transformation, I feel like they are unnecessary here.

For the latter (check and add to `var_lengths`), I think a better thing to do now is to throw a informative error message for now. For the former, could you give it a try to remove the returning of `final_param_idx`, if you couldn't make it work, we can leave it as is.

And for future reference, when I ask for a failure case, I'd love to see a concrete case where the program actually fail (preferably with code example), not just "this kind of programs might fail". This is only because intuition might not be accurate.

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -261,3 +259,194 @@ function evaluate_with_values(bn::BayesianNetwork, parameter_values::AbstractVec
...
+        # Compute log probability and continue
+        dist_logp = logpdf(dist, value) + logjac
+        remaining_logp = _marginalize_recursive(
+            bn, env, @view(remaining_nodes[2:end]), parameter_values, param_idx, var_lengths
+        )
+
+        return dist_logp + remaining_logp
+    end
+end
+"""
```
sunxd3, line 427:
```suggestion

"""
```

src/experimental/ProbabilisticGraphicalModels/bayesian_network.jl
```diff
@@ -261,3 +259,194 @@ function evaluate_with_values(bn::BayesianNetwork, parameter_values::AbstractVec
...
+
+    if isempty(discrete_vars)
+        return evaluate_with_values(bn, parameter_values)
+    end
+
+    # Initialize environment once
+    env = deepcopy(bn.evaluation_env)
+
+    # Initialize parameter index tracker
+    param_idx = Ref(1)
```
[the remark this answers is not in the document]
naseweisssss: Yes you are right, its unneccessary

Issue #281, opened 2025-03-04
Turing-like `@model` macro for specifying BUGS models


yebai:
[Logistic Regression with Random Effects](https://turinglang.org/JuliaBUGS.jl/stable/example/#Example:-Logistic-Regression-with-Random-Effects) using current syntax

```julia
model_def = @bugs begin
    for i in 1:N
        r[i] ~ dbin(p[i], n[i])
        b[i] ~ dnorm(0.0, tau)
        p[i] = logistic(alpha0 + alpha1 * x1[i] + alpha2 * x2[i] + alpha12 * x1[i] * x2[i] + b[i])
    end
    alpha0 ~ dnorm(0.0, 1.0E-6)
    alpha1 ~ dnorm(0.0, 1.0E-6)
    alpha2 ~ dnorm(0.0, 1.0E-6)
    alpha12 ~ dnorm(0.0, 1.0E-6)
    tau ~ dgamma(0.001, 0.001)
    sigma = 1 / sqrt(tau)
end
```

Desired Turing-like modelling syntax

```julia
JuliaBUGS.@model function model() begin
    for i in 1:N
        r[i] ~ dbin(p[i], n[i])
        b[i] ~ dnorm(0.0, tau)
        p[i] = logistic(alpha0 + alpha1 * x1[i] + alpha2 * x2[i] + alpha12 * x1[i] * x2[i] + b[i])
    end
    alpha0 ~ dnorm(0.0, 1.0E-6)
    alpha1 ~ dnorm(0.0, 1.0E-6)
    alpha2 ~ dnorm(0.0, 1.0E-6)
    alpha12 ~ dnorm(0.0, 1.0E-6)
    tau ~ dgamma(0.001, 0.001)
    sigma = 1 / sqrt(tau)
end
```

yebai:
Can we re-use `Turing.Gibbs` with this new unified syntax? That would be a nice feature. 

Related #185

sunxd3:
Not right now, unfortunately. Some reasons I can think of now:
* `Turing.Gibbs` is largely implemented using `GibbsContext`, so not entirely using `condition`/`deconsition`
* BUGS' `condition` right now actually creates "sub-model", not just conditioning. Not that this is a total block, but it needs some thoughts.

yebai:
@sunxd3, can you open an issue for Turing.jl on how to improve `Turing.Gibb` in a future meeting.

sunxd3:
[...]

PR #291, opened 2025-04-07
Implement Turing-like `model` macro

sunxd3:
Address https://github.com/TuringLang/JuliaBUGS.jl/issues/281

Edited:
I wrote some documentation in markdown for this PR and design, there is a slight change from last weeks' Monday meeting.

The markdown doc should be concise enough and I don't think I can make it much shorter without losing essential information.

So could the reviews, please, give https://github.com/TuringLang/JuliaBUGS.jl/blob/sunxd/new_model_syntax/docs/src/julia_syntax.md a quick read?


test/model_macro.jl
```diff
@@ -0,0 +1,83 @@
...
+        alpha0 ~ dnorm(0.0, 1.0E-6)
+        alpha1 ~ dnorm(0.0, 1.0E-6)
+        alpha2 ~ dnorm(0.0, 1.0E-6)
+        alpha12 ~ dnorm(0.0, 1.0E-6)
+        tau ~ dgamma(0.001, 0.001)
+        sigma = 1 / sqrt(tau)
+    end
+    #! format: on
+
+    @test_throws ErrorException begin
```
mhauru, line 35:
```suggestion
    # Try destructuring the random variables but forgetting to include one (tau).
    @test_throws ErrorException begin
```

test/model_macro.jl
```diff
@@ -0,0 +1,83 @@
...
+            alpha1 ~ dnorm(0.0, 1.0E-6)
+            alpha2 ~ dnorm(0.0, 1.0E-6)
+            alpha12 ~ dnorm(0.0, 1.0E-6)
+            tau ~ dgamma(0.001, 0.001)
+            sigma = 1 / sqrt(tau)
+        end
+        #! format: on
+    end
+
+    @test_throws ErrorException begin
```
mhauru, line 57:
```suggestion
    # Try leaving out one constant variable.
    @test_throws ErrorException begin
```

mhauru (commented):
I read the Markdown, it's well written. There's a broader conversation about `@parameters` that concerns DPPL too, as discussed e.g. today, but I assume that's outside the scope of this PR. I didn't actually read the implementation of the macros.

I added a couple of suggested comments to explain the tests, do edit them if I they are wrong.

PR #301, opened 2025-05-23
Clean Up Implementation for DP

naseweisssss:


Project.toml
```diff
@@ -7,6 +7,7 @@ AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 AbstractPPL = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
 Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
```
sunxd3, line 10: Could you remove these dependencies is they are not necessary for the code?

test/experimental/ProbabilisticGraphicalModels/bayesnet.jl
```diff
@@ -1,1225 +1,1576 @@
...
+            # Verify results match
+            @test isapprox(logp1, logp2, rtol=1e-10)
+
+            # Make sure there was at least some memoization
+            @test memo_size > 0
+
+            # For a chain of length 5, we could have at most 2*2^n_chain states
+            # The factor of 2 accounts for the fact that each node can be visited 
+            # once per possible configuration of its ancestors
+            @test memo_size <= 2 * 2^n_chain
```
sunxd3, line 1384: I am fine with this, although it feels like `memo_size` should be deterministic? This also applies to the tricky dependency graph test.

sunxd3 (commented):
It looks good to me in general, couple of small fixes

PR #307, opened 2025-05-30
Some minor tweaks 

yebai:
The scope of `interface.jl`  is unclear—it contains a set of miscellaneous functions defined for models, but it's hard to see why they differ from `condition` and `evaluation`. 

Update: merged `interface.jl` into `model.jl` to balance code per file and keep related code together.


sunxd3:
good changes!

what do you think instead of renaming `model_operations.jl` to `condition.jl`, but `abstractppl.jl`. My intention was that this file should contain all the interfaces of `abstractppl` including `condition/decondition`, `fix/unfix`, `evaluate!!` etc.

sunxd3:
Some other changes I plan to make is to rename the `evaluate!!` function with some names that are more clear what they are doing. Like `_ancestral_sampling`, and `AbstractPPL.evaluate!!` will just call these internal functions.

yebai:
All sounds good to me

sunxd3:
I have to rename `model.jl` to `bugsmodel.jl` to avoid name clash, otherwise I think this is in general better

PR #311, opened 2025-06-01
Refactor evaluation functions

sunxd3:


sunxd3:
some of the ideas are good, but I think this introduce quite a bit of complexity, in addition to being hard to make performant. So I'm not going to pursue this, but instead improve the current evaluation functions.

PR #314, opened 2025-06-02
Refactor evaluation functions

sunxd3:


src/model/abstractppl.jl
```diff
@@ -506,3 +506,78 @@ function _mark_as_unobserved(g::BUGSGraph, vars::Vector{<:VarName})
...
+"""
+    AbstractPPL.evaluate!!(rng::Random.AbstractRNG, model::BUGSModel; sample_all=true)
+
+Evaluate model using ancestral sampling from the given RNG.
+
+# Arguments
+- `rng`: Random number generator for sampling
+- `model`: The BUGSModel to evaluate
+- `sample_all`: If true, sample all variables; if false, only sample unobserved variables
+
```
mhauru, line 519:
```suggestion
- `temperature`: Temperature for tempering the likelihood (default 1.0)
- `transformed`: Whether to compute log density in transformed space (default model.transformed)

```

src/model/evaluation.jl
```diff
@@ -1,87 +1,204 @@
...
+- `temperature`: Temperature for tempering the likelihood (default 1.0)
+- `transformed`: Whether to compute log density in transformed space (default true)
+
+# Returns
+- `evaluation_env`: Updated evaluation environment
+- `(logprior, loglikelihood, tempered_logjoint)`: NamedTuple of log densities
+"""
+function evaluate_with_env!!(
+    model::BUGSModel,
+    evaluation_env=deepcopy(model.evaluation_env);
```
mhauru, line 129: Are you sure you need a deepcopy? A lot of people use it when they want a "safe" copy, but it's often considered bad practice: [[1](https://github.com/JuliaLang/julia/issues/42796#issuecomment-951232853), [2](https://ericphanson.com/blog/2024/why-you-might-avoid-deepcopy-in-julia/)]

test/model/evaluation.jl
```diff
@@ -1,4 +1,37 @@
-@testset "controlling sampling behavior for conditioned variables" begin
+function _logjoint(model::JuliaBUGS.BUGSModel)
+    _, log_densities = JuliaBUGS.evaluate_with_env!!(model; transformed=model.transformed)
+    return log_densities.tempered_logjoint
+end
+
+function verify_log_densities_structure(log_densities)
+    @test isa(log_densities, NamedTuple)
```
mhauru, line 7:
```suggestion
    @test isa(log_densities, NamedTuple)
    @test length(log_densities) == 3
```
Getting quite nitpicky here, but if we are checking the keys why not check that there are no other keys.

test/model/evaluation.jl
```diff
@@ -1,4 +1,37 @@
...
+
+function test_bugs_model_log_density(
+    model_example, expected_untransformed, expected_transformed=expected_untransformed
+)
+    (; model_def, data, inits) = model_example
+    transformed_model = compile(model_def, data, inits)
+    untransformed_model = JuliaBUGS.settrans(transformed_model, false)
+
+    # Allow world age to advance by calling the functions in a separate evaluation
+    Base.invokelatest() do
```
mhauru, line 21: Why is this needed? Asking just because I want to learn to understand this world age stuff.

mhauru (changes requested):
Looks, just a couple of small points and a question to satisfy my own curiosity.

test/model/evaluation.jl
sunxd3: In JuliaBUGS, I take the AST of the model definition and generate a function for each variable. These function are basically the RHS in DynamicPPL. 

I then call individual functions as I walk through graph nodes.

I create these functions by `eval`, so there are world-age problems. Hence I need to use `invokelatest`

src/model/evaluation.jl
sunxd3: This is one of things on my TODO list!

I traditionally use it for safety reasons, but I am moving away from it because with BUGS models, I actually know what is being mutated.

sunxd3:
(let me fix the test first. sorry)

mhauru (approved):

Issue #324, opened 2025-06-10
`of` for specifying parameter type


yebai:
Something like  `Julia.@parameters` but for more "dynamically" defining RVs instead of a struct block. 

```julia
# usecase 1
x = of(Array, 8)
y = of(Array, 4, 3)
y = of(Array, of(Real), 4, 3)
w = of(Real, lower, upper)

of((x, y, z, w)) # Tuple
Tparam = of((x=x,y=y,z=z, w=w)) # NamedTuple

model(rand(Tparam), constant1, constant2)


# usecase 2: type annotation

## tuple 
Tparameters = Tuple{of(Array, 8), of(Array, of(Real), 4, 3) of(Real, lower, upper)}

@model function demo(
	(x, y, w)::Tparameters,
	constants1, 
	constants2
	) 
end 

## namedtuple, @namedtuple(x::Tx, y::Ty) instead of @parameters(x::Tx, y::Ty) 

NTparameters = @NamedTuple{x::of(Array, 8), y::of(Array, of(Real), 4, 3) w::of(Real, lower, upper)}

@model function demo(
	(;x, y, w)::NTparameters,
	constants1, 
	constants2
	) 
end 

## direct support in `@model`

### tuple
@model function demo(
	(x::of(Array, 8), y::of(Array, of(Real), 4, 3) w::of(Real, lower, upper)),
	constants1, 
	constants2
	) 
end 

### namedtuple
@model function demo(
	(;x::of(Array, 8), y::of(Array, of(Real), 4, 3) w::of(Real, lower, upper)),
	constants1, 
	constants2
	) 
end 

# we can also do

x = zero(of(Array, 8)) 
x = rand(of(Array, 8))
x = of(Array, 3, 4)() # default to `zero(of(Array, 3, 4))` 

```

yebai:
Probably better to remove support for Tuple: 

```julia
# Tuple syntax 1
@model function demo(
	(x, y, w)::Tparameters,
	constants1, 
	constants2
	) 
end 

# Tuple syntax 2
@model function demo(
	(x::of(Array, 8), y::of(Array, of(Real), 4, 3), w::of(Real, lower, upper)),
	constants1, 
	constants2
	) 
end 
```

Since the above example is inconsistent with destructuring syntax, it would be unclear. On the other hand, support for NamedTupe is intuitive and clear: 

```julia

[...]

PR #331, opened 2025-06-13
Implement `of`

sunxd3:
address https://github.com/TuringLang/JuliaBUGS.jl/issues/324


DESIGN_of_proposal.md
```diff
@@ -0,0 +1,354 @@
...
+### Type Hierarchy
+
+```julia
+abstract type OfType end
+abstract type OfLeaf <: OfType end      # Terminal nodes
+abstract type OfContainer <: OfType end  # Container nodes
+
+# Leaf types (terminal nodes)
+struct OfArray{T,N} <: OfLeaf
+    element_type::Type
```
yebai, line 110:
```suggestion
    element_type::T
```

DESIGN_of_proposal.md
```diff
@@ -0,0 +1,354 @@
...
+@model function demo((;x, y, w)::NTparameters, constants...)
+    # model body
+end
+```
+
+#### Use Case 3: Direct Support in `@model`
+```julia
+# Tuple syntax
+@model function demo(
+    (x::of(Array, 8), y::of(Array, of(Real), 4, 3), w::of(Real, lower, upper)),
```
yebai, line 84: One issue with such tuple syntax is variables `x`, `y`, etc won't be available in the function body since Julia sees the first argument `(x, y, ...)` as a single argument by default. We can override this behaviour, but that might surprise users.

DESIGN_of_proposal.md
```diff
@@ -0,0 +1,354 @@
...
+        )), 
+        10  # 10 groups
+    ),
+    # Individual-level parameters
+    sigma = of(Real, 0, nothing)          # Observation noise
+))
+
+# Use in model
+@model function hierarchical_model(params::HLMParams, X, y, group_id)
+    (; population, groups, sigma) = params
```
yebai, line 218:
```suggestion
@model function hierarchical_model((; population, groups, sigma)::HLMParams, X, y, group_id)
```

DESIGN_of_proposal.md
```diff
@@ -0,0 +1,354 @@
...
+            mu = of(Array, 2),              # 2D mean
+            precision = of(Array, 2, 2)     # Precision matrix
+        )),
+        20  # Maximum 20 components
+    ),
+    assignments = of(Array, Int, 1000)      # Cluster assignments
+))
+
+@model function gmm_model(params::GMMParams, data)
+    (; K, weights, components, assignments) = params
```
yebai, line 303:
```suggestion
@model function gmm_model((; K, weights, components, assignments)::GMMParams, data)
```

src/model_macro.jl
```diff
@@ -118,14 +118,33 @@ function _generate_model_definition(model_function_expr, __source__, __module__)
...
+        # Tuple with type annotation: (x, y, z)::ParamType
+        param_type = ptype
+        param_fields = extract_field_names(fields)
+    elseif MacroTools.@capture(param_destructure, (fields__,))
+        # Tuple without type annotation: (x, y, z) or (x::of(...), y::of(...))
+        param_fields, param_annotations = extract_fields_with_annotations(fields)
+    else
+        return :(throw(
+            ArgumentError(
+                "The first argument of the model function must be a destructuring assignment (tuple or named tuple).",
```
yebai, line 144: Support for Tuple needs discussion. See comment above.

test/test_of_type.jl
```diff
@@ -0,0 +1,301 @@
...
+    JuliaBUGS.@model function test_model2((; x, y), n)
+        for i in 1:n
+            x[i] ~ Normal(0, 1)
+        end
+        return y ~ Uniform(0, 1)
+    end
+
+    # Test model instantiation
+    params = of((x=of(Array, 3), y=of(Real, 0.0, 1.0)))
+    model1 = test_model1(rand(params), 3)
```
yebai, line 296:
```suggestion
    Tparams = of((x=of(Array, 3), y=of(Real, 0.0, 1.0)))
    model1 = test_model1(rand(Tparams), 3)
```

yebai (changes requested):
I took a quick look and left a few minor comments.

sunxd3:
@yebai I just pushed some local changes. Can you give it another quick read?

https://github.com/TuringLang/JuliaBUGS.jl/blob/sunxd/of/of_design_doc.md

yebai:
A few comments: 


1. Support `dimention(of_type)` to retrieve the dimension of `of` types or instances
2. Replace`validated_params = validate(Tparams, user_input)` with `Tparams(user_input)`, i.e. use a constructor instead of `validate` API. The namedtuple-based constructor is more intuitive. 
3. I don't see the motivation of the `@of` macro. All we need is already captured by `of()`, which has all the type information for defining a new type if required. 
4. All LHS variables of tilde should be from the destructuring of the first input argument. 
```julia

# always do 
@model function school_model((; mu0, beta, tau2, sigma2, school_effects)::SchoolParams, data)
    (; y, X, school_id, n_students) = data
end

# instead of 
@model function school_model(params::SchoolParams, data)
    (; mu0, beta, tau2, sigma2, school_effects) = params
    (; y, X, school_id, n_students) = data
end
```

The motivation is that we want to enforce that all LHS of tilde has to be a variable from the destructured first argument of models (e.g. `mu0`, `beta`, etc in `school_model((; mu0, beta, tau2, sigma2, school_effects)::SchoolParams, ...)` later. This will help us resolve the long-standing [issue](https://github.com/TuringLang/DynamicPPL.jl/issues/519). In the above new syntax, a variable from the first argument doesn't tell whether it is a parameter or data. That is, the "observation" step will be decoupled and performed using the `condition`/`decondition` functions. This is slightly different from the current behaviour in `DynamicPPL.model` that input arguments on tilde's LHS will always be data, which will be deprecated.

sunxd3:
there are some issues
* to support something like `x::of(Real, ...)` of need to return a `Type` object. the most straightforward thing to do is to return `Real` (as in `of(Real) == Real`) the issue is that `Real` doesn't carry extra information like lower and upper bounds
* I added the macro `@of` to be able to use it on the RHS of `::`

sunxd3:
@yebai comments at https://github.com/TuringLang/JuliaBUGS.jl/pull/331#issuecomment-2976087364 are addressed.

couple of deviations:
1. use `size` and `length` instead of `dimension`, we can bind `dimension` to `length`, of course
2. `Tp = of((;n=of(Constant), m=of(Constant), x=of(Array,n,m)))` is hard to make work, I opted for `Tp = of((;n=of(Constant), m=of(Constant), x=of(Array,:n,:m)))`

The updated design doc is at https://github.com/TuringLang/JuliaBUGS.jl/blob/sunxd/of/of_design_doc.md and code is at https://github.com/TuringLang/JuliaBUGS.jl/blob/sunxd/of/src/of_type.jl. The source code required no dependencies, all the example code in the design doc should run, except for the model macro (I haven't made substantial updates to JuliaBUGS.@model` yet to reflect `of`.

sunxd3:
https://github.com/TuringLang/JuliaBUGS.jl/issues/324#issuecomment-2980915683 is mostly implemented. 

The part left over is support of expressions in array size. I.e., currently
```julia
Tparams = @of(
   b=of(Int), 
   θ=of(Array, b, 1), 
   y=of(Real)
)
```
is fine, but 
```julia
Tparams = @of(
   b=of(Int), 
   θ=of(Array, b+1, 1), 
   y=of(Real)
)
```
is not. 

Two things to consider further:
- Julia constraints what can be used as type parameter
```julia
julia> A{:(x+1)}
ERROR: TypeError: in Type, in parameter, expected Type, got a value of type Expr
Stacktrace:
 [1] top-level scope
   @ REPL[2]:1

julia> A{:x}
A{:x} 
```
- what functions to allow in these size expressions

of_design_doc.md
```diff
@@ -151,3 +179,81 @@ new_params = unflatten(ParamsType, new_flat_params)
     end
 end
 ```
+
+### Dynamic Model with Variable Dimensions
```
yebai, line 183:
```suggestion
### Model with Variable Dimensions
```

of_design_doc.md
```diff
@@ -151,3 +179,81 @@ new_params = unflatten(ParamsType, new_flat_params)
...
+# Concrete type with specific dimensions
+ConcreteARType = of(ARParams; order=3)
+# This creates a type where coeffs has size 3
+
+data = (
+    coeffs=[0.5, -0.3, 0.1],
+    sigma=0.25,
+    y=randn(100)
+)
+params = ConcreteARType(data)
```
yebai, line 258:
```suggestion
params = ConcreteARType(;data=data)
```

of_design_doc.md
```diff
@@ -151,3 +179,81 @@ new_params = unflatten(ParamsType, new_flat_params)
     end
 end
 ```
+
+### Dynamic Model with Variable Dimensions
+
+Here's an example of a model where array dimensions depend on other parameters:
+
+```julia
+# Model where the number of mixture components is dynamic
```
yebai, line 188:
```suggestion
# Mixture model where the number of mixture components is itself a parameter
```

of_design_doc.md
```diff
@@ -151,3 +179,81 @@ new_params = unflatten(ParamsType, new_flat_params)
...
+Tparams = @of(
+    n_components=of(Int, 1, 3; constant=true),  # Can be 1, 2, or 3
+    weights=of(Array, n_components),            # Size depends on n_components
+    means=of(Array, n_components),              # Size depends on n_components
+    y=of(Real)
```
yebai, line 193: Let's have `N` data points `y = of(Array, N, 1)`

of_design_doc.md
```diff
@@ -151,3 +179,81 @@ new_params = unflatten(ParamsType, new_flat_params)
...
+
+# Another example: Variable-order autoregressive model
+ARParams = @of(
+    order=of(Int, 1, 5; constant=true),  # AR order between 1 and 5
+    coeffs=of(Array, order),             # AR coefficients
+    sigma=of(Real, 0, nothing),          # Error variance
+    y=of(Array, 100),                    # Time series data
+)
+
+@model function dynamic_ar((; order, coeffs, sigma, y)::ARParams, n_obs)
```
yebai, line 230:
```suggestion
@model function variable-order_ar((; order, coeffs, sigma, y)::ARParams, n_obs)
```

of_design_doc.md
```diff
@@ -151,3 +179,81 @@ new_params = unflatten(ParamsType, new_flat_params)
...
+        coeffs[i] ~ Normal(0.0, 0.5)  # Shrinkage prior on AR coefficients
+    end
+    sigma ~ InverseGamma(2.0, 1.0)
+    
+    # AR likelihood
+    for t in (order+1):n_obs
+        # Compute AR prediction
+        pred = 0.0
+        for i in 1:order
+            pred += coeffs[i] * y[t-i]
```
yebai, line 242: This aligns better with the notation `coeffs' * y` 
```suggestion
            pred += coeffs[order+1-i] * y[t-i]
```

of_design_doc.md
```diff
@@ -39,33 +41,59 @@ The system encodes extra useful information into type parameters:
...
+)
+```
+
+### 5. Symbolic Dimensions and Bounds
 
 For cases where dimensions need to be specified at runtime:
 
 ```julia
 # Define type with symbolic dimensions
 MatrixType = of((
```
yebai, line 68: I suggest we remove support for using `of((;...))` and only keep `@of()` for composable types for simplicity and to reduce the maintenance burden.

yebai (changes requested):

yebai:
Good work @sunxd3 -- I left a few more comments above. We are getting there. 

> what functions to allow in these size expressions

For now, let's only support `*`, `+`, `-`, and `/`.

of_design_doc.md
```diff
@@ -64,35 +58,29 @@ T = @of(
...
-ConcreteType = of(MatrixType; rows=3, cols=4)
-data = (data=rand(3, 4),)
-ConcreteType(data)
-flat = flatten(ConcreteType, data)
+# Create concrete type by specifying constants
+ConcreteType = MatrixType(;rows=3, cols=4)
+# Can also validate data while concretizing
+ConcreteType = MatrixType(;rows=3, cols=4, data=rand(3, 4))  # Validates that data matches 3x4
+data_nt = (data=rand(3, 4),)
+flat = flatten(ConcreteType, data_nt)
```
yebai, line 74: Note `ConcreteType` already has randomly initialised `data`, so `data_nt` is unnecessary. 
```suggestion
flat = flatten(ConcreteType)
```

sunxd3: this doesn't feel quite right -- `ConcreteType` in your example should be a type, i.e. 
```julia
ConcreteType == @of(
data = of(Array, 3, 4)
)
```
`flatten` should take a source of values that conforms to the `of` type and returns a vector.

sunxd3: at the moment, as of the last commit

```julia
ConcreteType = MatrixType(;rows=3, cols=4, data=rand(3, 4))  # Validates that data matches 3x4
```
will only check that the size of `data` is consistent with
```julia
MatrixType = @of(
    rows=of(Int; constant=true),
    cols=of(Int; constant=true),
    data=of(Array, rows, cols),  # Direct references converted to symbols
)
```
with
 `rows=3, cols=4`

yebai: > will only check that the size of data is consistent 

That would be confusing. I suggest

```julia
x = MatrixType(;rows=3)  # this should error with a message constant `cols` is required but not provided. 
x = MatrixType(;rows=3, cols=4)  # default to MatrixType(;rows=3, cols=4, data = zero(3, 4))
x = MatrixType(;rows=3, cols=4, data = data) # 
```

This would always return an instance when calling `T(;kwargs...)` instead of another type. 
In addition, it requires the user to pass all constants when calling `T(;kwargs...)`, and default to zero if non-constant variables aren't provided explicitly.

sunxd3: this is updated now

sunxd3:
design at the moment: https://github.com/TuringLang/JuliaBUGS.jl/blob/sunxd/of/of_design_doc.md

image

use of expressions is also supported (an example is added at https://github.com/TuringLang/JuliaBUGS.jl/blob/sunxd/of/of_design_doc.md#bayesian-nonparametric-clustering-model)

of_design_doc.md
```diff
@@ -66,21 +66,25 @@ MatrixType = @of(
...
+# Create instance with specific data
+instance = MatrixType(;rows=3, cols=4, data=rand(3, 4))  
+# instance = (data = <provided 3x4 matrix>,)
 
-# rand and zero will first concretize the type and call rand
-rand(MatrixType(; rows=3, cols=4))
-zero(MatrixType(; rows=10, cols=5))
+# Use flatten/unflatten with unconcretized types (providing constants as kwargs)
+flat = flatten(MatrixType, instance; rows=3, cols=4)
+reconstructed = unflatten(MatrixType, flat; rows=3, cols=4)
```
yebai, line 79: We can use `of(MatrixType::OfType; rows=3, cols=4)` for creating semi-concrete types, where `rows=3, cols=4` serves a similar role to `lower` and `upper`. 

```suggestion
reconstructed = unflatten(of(MatrixType; rows=3, cols=4), flat; rows=3, cols=4)
```

Note that the first argument of this syntax has to be a subtype of `OfRype`.

of_design_doc.md
```diff
@@ -66,21 +66,25 @@ MatrixType = @of(
...
 
-rand(MatrixType(; rows=3)) # this would fail because some `Constant`s are not specified
+# rand and zero with keyword arguments
+rand(MatrixType; rows=3, cols=4)  # generates random instance
+zero(MatrixType; rows=10, cols=5) # generates zero instance
```
yebai, line 83:
```suggestion
rand(of(MatrixType; rows=3, cols=4))  # generates random instance
zero(of(MatrixType; rows=3, cols=4))  # generates zero instance
```

of_design_doc.md
```diff
@@ -66,21 +66,25 @@ MatrixType = @of(
...
+
+# Missing constants will error
+MatrixType(; rows=3) # Error: Constant `cols` is required but not provided
+rand(MatrixType; rows=3) # Error: Missing values for symbolic dimensions: cols
```
yebai, line 87:
```suggestion
rand(of(MatrixType; rows=3)) # Error: Missing values for symbolic dimensions: cols
```

of_design_doc.md
```diff
@@ -92,14 +96,16 @@ ExpandedMatrixType = @of(
...
+instance = ExpandedMatrixType(; n=10)
+# This creates an instance with:
+# - original: 10×10 zero matrix
+# - padded: 11×11 zero matrix
+# - doubled: 20×10 zero matrix  
+# - halved: 5×10 zero matrix  (n/2 must result in an integer, error if not)
 
-rand(ConcreteExpanded)
+# Generate random instance
+rand(ExpandedMatrixType; n=10)
```
yebai, line 108:
```suggestion
rand(of(ExpandedMatrixType; n=10))
```

yebai (commented):
Good work -- I added a few final comments on constructor syntax, mainly for consistency.

of_design_doc.md
sunxd3: yeah, this is already implemented, I can update the design doc

yebai:
Excellent -- we now have a complete `of`-type design!

of_design_doc.md
yebai: @sunxd3 For clarity, let's make sure to remove these methods

```julia
zero(ExpandedMatrixType; n=10)
rand(ExpandedMatrixType; n=10)
flat = latten(MatrixType, instance; rows=3, cols=4)
unflatten(MatrixType, flat; rows=3, cols=4)
```

in favour of

```julia
zero(of(ExpandedMatrixType; n=10))
rand(of(ExpandedMatrixType; n=10))
flat = latten(of(MatrixType, instance; rows=3, cols=4))
unflatten(of(MatrixType, flat; rows=3, cols=4))
```

for the principle that there should be only one way of doing things.

sunxd3: done now

docs/src/julia_syntax.md
```diff
@@ -160,28 +209,40 @@ BUGSModel (parameters are in transformed (unconstrained) space, with dimension 4
...
+        end
+        alpha0 ~ dnorm(0.0, 1.0E-6)
+        alpha1 ~ dnorm(0.0, 1.0E-6)
+        alpha2 ~ dnorm(0.0, 1.0E-6)
+        alpha12 ~ dnorm(0.0, 1.0E-6)
+        tau ~ dgamma(0.001, 0.001)
+        sigma = 1 / sqrt(tau)
+    end
+
+julia> m2 = seeds_v2(NamedTuple(), x1, x2, N, n)  # Empty NamedTuple for no observations
```
yebai, line 247:
```suggestion
julia> m2 = seeds_v2((;), x1, x2, N, n) # Empty NamedTuple for no observations
```

yebai (commented):
Added a few comments following today's discussion. Specifically, added a new syntax suggestion for optional type annotation in inlined parameter type syntax.

src/of_type.jl
```diff
@@ -790,6 +790,72 @@ function Base.zero(::Type{OfConstantWrapper{T}}) where {T}
...
+end
+
+function Base.missing(::Type{OfConstantWrapper{T}}) where {T}
+    return error(
+        "Cannot generate missing values for constants. Use missing(of(T; const_name=value)) after providing the constant value.",
+    )
+end
+
+# Create instance with missing values from concretized of type
+function Base.missing(T::Type{<:OfType})
```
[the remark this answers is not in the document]
sunxd3: I opted for `unflatten` for now, this doesn't feel perfect, but I don't have an idea that's much better. We can iterate in the future.

yebai:
@penelopeysm / @mhauru, can any of you give this a look before we merge it? I have provided extensive comments throughout the process.

of_design_doc.md
```diff
@@ -0,0 +1,334 @@
...
+- **Dimensions**: Stored as tuple type parameters (e.g., `(3, 4)` for a 3×4 matrix)
+- **Bounds**: Encoded using `Val{x}` for numeric bounds or `Nothing` for unbounded
+- **Symbolic references**: Encoded using `SymbolicRef{:symbol}` for referencing other fields
+- **Arithmetic expressions**: Encoded using `SymbolicExpr{expr}` for expressions like `n+1`, `2*n`, etc. Division operations must result in integers for array dimensions.
+- **Field names**: Stored as tuple of symbols in `OfNamedTuple`
+- **Element types**: Preserved as type parameters for arrays and nested structures
+
+### 3. Operations on Types
+
+- `T(;kwargs...)` where `T<:OfType` - Create instances with specified constants (returns values, not types)
```
yebai, line 36:
```suggestion
- `T(default_value=0.;kwargs...)` where `T<:OfType` - Create instances with specified constants (returns values, not types), and initialise all element values to `default_value`, e.g. `T(default_value=0.;kwargs...)` initialise all element values to `missing`. 
```

docs/src/julia_syntax.md
```diff
@@ -160,28 +209,58 @@ BUGSModel (parameters are in transformed (unconstrained) space, with dimension 4
...
-  alpha2  = -545.664
-  alpha12 = 1338.25
-  tau     = 0.0
+julia> # Create model with observations for r
+julia> r_data = [10, 23, 23, 26, 17, 5, 53, 55, 32, 46, 10, 8, 10, 8, 23, 0, 3, 22, 15, 32, 3]
+julia> m_obs = seeds((r=r_data,), x1, x2, N, n)
+
+julia> # Using unflatten for initialization with missing values
+julia> using JuliaBUGS: unflatten
+julia> params = unflatten(of(SeedsParams), missing)
```
mhauru, line 218: Why do I still need `of(SeedsParams)` when `SeedsParams` was created with `@of`?

docs/src/julia_syntax.md
```diff
@@ -75,8 +75,27 @@ JuliaBUGS provides a Julian interface inspired by Turing.jl's model macro syntax
 
 ### The `@model` Macro
 
+The `@model` macro provides a Julia-native interface for defining probabilistic models using external type definitions.
+
+#### External Type Definition with `@of`
```
mhauru, line 80: I'm not very keen on the name `of`. It's creating X of Y, but it's not clear from context or the name what X and Y can be. For instance, in some languages, `of` is used in names of functions that convert from one type to another, as in `String.of_int` (from OCaml) that turns an int to a string, but that's not what's happening here.

Would something like `parameter_type` work? I know its more verbose, but it would make it much more concrete and obvious what the function/macro is doing.

src/of_type.jl
```diff
@@ -0,0 +1,1602 @@
...
+Create an OfNamedTuple type with cleaner syntax that allows direct references
+between fields without needing symbols.
+
+# Examples
+```julia
+T = @of(
+    rows=of(Int; constant=true),
+    cols=of(Int; constant=true),
+    data=of(Array, rows, cols)
+)
```
mhauru, line 594: Does order matter here? Could `data` come before `rows` and `cols`? I assume `rows` and `cols` have to be `constant=true` for them to be used in creating `data`? Would be good to document these, and then probably stick the doctsring of `@of` and `of` in the docs as well.

src/of_type.jl
```diff
@@ -0,0 +1,1602 @@
...
+    else
+        bound_to_type(upper)
+    end
+    return L, U
+end
+
+# ========================================================================
+# Constructor Functions
+# ========================================================================
+function of(::Type{Array}, dims...; constant::Bool=false)
```
mhauru, line 219: Could `of` have a docstring and could it be added to the docs? It's a bit unclear to me what the general semantics of `of` are, and what exactly it returns.

src/of_type.jl
```diff
@@ -0,0 +1,1602 @@
+using Random: randexp
+
+# ========================================================================
+# Core Type Definitions
+# ========================================================================
+
+# Abstract base type for all Of types
+abstract type OfType end
```
mhauru, line 8: Is this extendable, in the sense that if a user would want to add their own parameter type, can that be done? If yes, would it require knowing the internals of JuliaBUGS, or is there some interface to use?

Not saying it should be, just asking. I think for DPPL it should be, but this is not DPPL.

of_design_doc.md
```diff
@@ -0,0 +1,344 @@
...
+
+### 1. Type-Based Design
+
+The `of` function returns types with specifications encoded in type parameters:
+- `of(Array, dims...)` → `OfArray{Float64, N, (dim1, dim2, ...)}` - Arrays with specified dimensions
+- `of(Array, T, dims...)` → `OfArray{T, N, (dim1, dim2, ...)}` - Typed arrays
+- `of(Real)` → `OfReal{Nothing, Nothing}` - Unbounded real numbers
+- `of(Real, lower, upper)` → `OfReal{Val{lower}, Val{upper}}` - Bounded real numbers
+- `of(Int)` → `OfInt{Nothing, Nothing}` - Unbounded integers
+- `of(Int, lower, upper)` → `OfInt{Val{lower}, Val{upper}}` - Bounded integers
```
mhauru, line 20: Does this need the `Val`s? Bits types can be type parameters as they are.

of_design_doc.md
```diff
@@ -0,0 +1,344 @@
...
+- `@of(field1=..., field2=...)` → `OfNamedTuple{(:field1, :field2), Tuple{Type1, Type2}}` - Named tuples (use @of macro only)
+- `of(...; constant=true)` → `OfConstantWrapper{T}` - Marks a type as constant/hyperparameter (only supported for `Int` and `Real` types)
+
+### 2. Type Parameter Encoding
+
+The system encodes extra useful information into type parameters:
+- **Dimensions**: Stored as tuple type parameters (e.g., `(3, 4)` for a 3×4 matrix)
+- **Bounds**: Encoded using `Val{x}` for numeric bounds or `Nothing` for unbounded
+- **Symbolic references**: Encoded using `SymbolicRef{:symbol}` for referencing other fields
+- **Arithmetic expressions**: Encoded using `SymbolicExpr{expr}` for expressions like `n+1`, `2*n`, etc. Division operations must result in integers for array dimensions.
```
mhauru, line 30: After reading the examples I see why something like `SymbolicExpr` is needed. It makes me somewhat uncomfortable though, it feels like we are mixing source code (`Expr`s) with values in a way that at least at first seems confusing to me. I wonder if the `Expr`s could somehow be replaced with functions. This isn't so much a comment asking for change, but more a comment flagging something that I don't understand but that seems unusual and possibly troublesome to me.

docs/src/julia_syntax.md
```diff
@@ -91,34 +110,73 @@ JuliaBUGS.@model function model_definition((;r, b, alpha0, alpha1, alpha2, alpha
...
-4. The remaining arguments must specify all constants and independent variables required by the model (variables used on the RHS but not on the LHS).
+#### The `of` Type System
 
-The `@parameters` macro simplifies creating structs to hold model parameters:
+The `of` function creates type specifications for parameters:
+
+- `of(Real)` - Unbounded real number
+- `of(Real, lower, upper)` - Bounded real number
+- `of(Int)` - Unbounded integer  
+- `of(Int, lower, upper)` - Bounded integer
```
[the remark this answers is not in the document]
yebai: Happy to go with `Float64` / `Float32`.

src/of_type.jl
yebai: Yes, it should be extensible. The `of_type` will be moved to `AbstractPPL` later, so it won't depend on JuliaBUGS internals.

of_design_doc.md
penelopeysm: The problem with anon functions is that you can call the parameter `n` but it has no knowledge that that is supposed to be the `n` that came before it. These functions are the same: 

```julia
i -> i + 1
n -> n + 1
```

(cf. alpha conversion in lambda calculus), because the value of `n` isn't the `n` from outside but rather the parameter, so how can we tell whether it's supposed to add one to `i` or `n`? You need to carry a reference to a symbol that is resolved **outside of** the scope of the function. This can only be accomplished with some sort of `Expr` machinery.

mhauru: I'm not sure I understand. The macro knows the names of each of the fields, so it could write the anonymous function as `n -> n + 1` if the first field is called `n`. After macro processing is done, I agree that the distinction between that and `i -> i + 1` is lost, but these functions would always be called with the preceding fields as arguments, in order as positional arguments. So the call would be something like
```julia
constructed_value_for_field_4 = anonymous_functions_for_field_4(val_of_field1, val_of_field2, val_of_field3)
```
Wouldn't that do it?

penelopeysm: Firstly, sorry, I didn't see your second comment before I posted mine. When I posted it Hong's comment was the last one in the thread (I'm going to blame mobile data on the Tube for that).

> Wouldn't that do it?

Sure, it would, in that the _output_ of the macro would be an anon function. I had in mind that you all wanted the _input_ of the macro to be an anon function, e.g.

```julia
@of(
    halved=(n -> of(Array, n/2, n))
)
```

This wouldn't work for the reasons noted in my comment.

Sure, I think having the macro output anon functions instead of Exprs may be simpler. As you said, it does rely on specific ordering of variables which seems to go against the design suggested elsewhere.

Overall, the point I'm trying to make (and that seems to be lost) is that variable names like this inevitably introduce complexity that has to be dealt with at the Expr level.

There is a fundamental difference between `of(Real)` and `of(Array, n/2, n)`. The first one is an actual Julia function that will actually return something if used outside of the macro. Thus, the macro does not need to _interpret_ `of(Real)`, it can simply pass it through. The second one is inherently nonsensical and it is not possible to write (non-macro) Julia code to give it a meaning. It can only be understood within the context of the `@of` macro.

Regardless of whether the `@of` macro outputs a partially parsed Expr where references to `n` are resolved later, or whether it converts it into an anonymous function where references to `n` are implicitly resolved by `n` being the first parameter, it's still the same amount of complexity.

src/of_type.jl
sunxd3: some docstrings added now

sunxd3:
Addressed two of @mhauru's suggestions: 1. add docstrings 2. add support for `of(Float64)`.

src/of_type.jl
```diff
@@ -4,24 +4,113 @@ using Random: randexp
...
 abstract type OfType end
 
-# Wrapper type for symbolic references in bounds
+"""
+    SymbolicRef{S}
+
+Wrapper type for symbolic references in bounds and dimensions.
+
+Used internally to encode references to other fields when specifying bounds or dimensions 
+that depend on constant fields. For example, when using `@of(n=of(Int; constant=true),
```
yebai, line 35: Bounds or dimensions may also depend on other parameters. 
```suggestion
that depend on constant fields or other parameters. For example, when using `@of(n=of(Int; constant=true), 
```

src/of_type.jl
```diff
@@ -4,24 +4,113 @@ using Random: randexp
...
+"""
 struct SymbolicRef{S} end
 
-# Wrapper type for symbolic expressions (e.g., b+1, 2*n)
+"""
+    SymbolicExpr{E}
+
+Wrapper type for symbolic expressions in dimensions.
+
+Used internally to encode arithmetic expressions involving constant fields. For example,
```
yebai, line 52:
```suggestion
Used internally to encode arithmetic expressions involving constant fields or other parameters. For example,
```

src/of_type.jl
```diff
@@ -216,6 +370,76 @@ end
...
+
+## Integers
+```julia
+of(Int)                         # Unbounded integer
+of(Int, lower, upper)           # Bounded integer
+```
+
+## Named Tuples
+```julia
+of((field1=spec1, field2=spec2, ...))  # NamedTuple with typed fields
```
yebai, line 405:
```suggestion
of((;field1=spec1, field2=spec2, ...))  # NamedTuple with typed fields
```

Issue #312, opened 2025-06-02
Remove type piracy methods


sunxd3:
https://github.com/TuringLang/JuliaBUGS.jl/blob/9ccc193260e18f1871957995ec0deffbef6d70aa/src/model/utils.jl#L1-L22

ref: https://github.com/TuringLang/JuliaBUGS.jl/pull/309#discussion_r2120541889

PR #337, opened 2025-06-23
WIP: Address type piracy in array operations (tests currently failing)

arnavk23:
The current implementation extends BangBang.jl's methods directly (_setindex and setindex!!), which constitutes type piracy and may cause conflicts with other packages. While functional, this approach is discouraged in the Julia ecosystem. Fixes (#312)

- Encapsulated the operations in a SafeBangExtensions module to avoid global method extensions
- Implemented non-pirating versions (_safe_array_update and _namedtuple_update) with identical functionality


yebai:
Hi @arnavk23, JuliaBUGS is a scientific package and requires extensive knowledge about Julia and Bayesian inference. Please reach out to a team member if you would like to contribute regularly.

Issue #94, opened 2023-09-18
`AdvancedHMCExt` will use `evaluate!!` to rerun the model for every sample drawn, and it seems wasteful  


sunxd3:
See conversation https://github.com/TuringLang/JuliaBUGS.jl/pull/93#discussion_r1328675770

yebai:
@sunxd3, is this actually fixed?

sunxd3:
no, I should've added that this is superseded by https://github.com/TuringLang/JuliaBUGS.jl/issues/327

there are subtle differences, but a good solution of #327 will resolve this issue or makes it much easier

PR #338, opened 2025-06-26
AdvancedHMCExt

arnavk23:
AdvancedHMCExt will use evaluate!! to rerun the model for every sample drawn, and it seems wasteful  #94


sunxd3:
may I know what you're achieving with the changes?

arnavk23:
1. **Avoid Redundant `evaluate!!` Calls**

Previously, the module reran `evaluate!!` inside a custom `AbstractMCMC.step` method for every transition. This is **unnecessary**, because:

* `AdvancedHMC.step` already returns valid parameter values in `t.z.θ` that reflect the model's state.
* `evaluate!!` can be expensive, especially in large models.

By **removing the post-sampling evaluation**, we save compute time and memory. The model is re-evaluated only when needed (e.g., at Gibbs step boundaries), not every iteration.

2. **Use `getparams` Directly From Transitions**

Instead of re-evaluating the model to get updated parameters, the new code:

* Uses `t.z.θ` as the sampled parameter vector,
* Applies `invlink` to undo any transformations,
* And optionally uses the AD-wrapped log density's transform info to extract readable parameters.

This means `bundle_samples` now builds MCMC chains **without reinitializing the model** on each sample.

3. **Simplify Fallback `step` Method**

The fallback `AbstractMCMC.step(rng, model::AbstractBUGSModel, sampler::HMC)` was:

* Manually calling `evaluate!!` after sampling,
* Rebuilding `varinfo`, `context`, etc., even if not necessary.

Now, it just returns the new parameter dictionary (filtered by `should_save`), skipping unnecessary work unless diagnostics or re-evaluation are explicitly needed.

4. **Improve Error Handling for Missing AD Backend**

If a user tries to use an `HMC` or `NUTS` sampler **without specifying an AD backend**, they get a clear error:

```julia
Gradient-based samplers (HMC/NUTS) require an explicit AD backend.
```

arnavk23:
Sure! The main goal of this PR is to extend JuliaBUGS so it defines `update_sampler_state` to handle updating sampler states with gradients.
Along with that, I also added an experimental `mh_from_prior` implementation, which makes it possible to draw samples directly from the prior using a simple Metropolis-Hastings step.

arnavk23:
@sunxd Both of the failing checks are failing at bayesnet and I have not made any change to bayesnet (but it is showing in changed file).

PR #347, opened 2025-07-17
DoodleBUGS: Basic Code Generation, Advanced Exports, and State Persistence

shravanngoswamii:
Advanced Export Functionality:
- Added a `svg` and `jpg` exports.
- Added a ExportModal component that allows for customization of all image exports (background, scale, quality, etc.).
- All export options (PNG, JPG, SVG, JSON) are now under a single "Export" dropdown in navbar.

State Persistence:
- The application now remembers and restores the user's session on page reload, including the last active project, the last opened graph, custom sidebar widths, active tab in right sidebar.
- Sidebars are now resizable, with their widths persisted in localStorage.

Advanced Code Editor:
- Upgraded the JSON editor and BUGS code preview panel to use CodeMirror with syntax highlighting, line numbers, and better scrolling.
- fields (like id) are now marked as read-only within the editor, so user won't be able to edit it.

BUGS Code Generation:
- A new `useBugsCodeGenerator` composable now generates BUGS model code from the graph structure (Not the correct code but it's recognizing nodes and edges properly, still need too much work and fixes).
- Implemented a topological sort to ensure correct dependency order.

Centralized Configuration: Refactored the app to use a central `nodeDefinitions.ts` file as the single source of truth for all node types and properties.

Bug Fixes:
- Fixed a bug that caused `png` export to generate corrupted files.
- Fixed the right-click context menu `Remove` to correctly update the application state.
- Plates are now automatically deleted when they become empty.

TODO before merging this PR:
- [ ] Implement two-way highlighting: selecting an element on the canvas should scrolls to and highlight the corresponding object in the JSON editor. (I have tried to implement this but facing some issues)
- [ ] Data Input Panel: It's not static now but still does not work properly!

PR #353, opened 2025-07-25
DoodleBUGS: consistent look for DataInputPanel like JSON Editor and code preview 

AJ0070:
In support of #347 
cc @shravanngoswamii


shravanngoswamii:
Thank you @AJ0070!

PR #373, opened 2025-08-10
improve coverage GHA

shravanngoswamii:
`julia-actions/julia-processcoverage@v1` step logs are not finding files now!


shravanngoswamii:
@sunxd3 Can you please have a look at these failure, you have more context of these...

PR #378, opened 2025-08-14
Update JuliaBUGS/docs/Project.toml

shravanngoswamii:
Monorepo packages are prefixed with folder name so tag documentation build requires `tag_prefix` but it is not supported by old Documenter.jl versions.

https://github.com/TuringLang/JuliaBUGS.jl/blob/d032e614af5c2bfa8abf603ab8b03294754c5d1b/.github/workflows/Docs.yml#L41

Now the issues is that new Documenter.jl version is causing docs failure because we are calling Julia standard library functions in docs (I am not sure of this), I am assuming it looking at these functions
https://github.com/TuringLang/JuliaBUGS.jl/blob/d032e614af5c2bfa8abf603ab8b03294754c5d1b/JuliaBUGS/docs/src/api/functions.md?plain=1#L10C1-L31C1

### DIFF
--- a/.github/workflows/Docs.yml
+++ b/.github/workflows/Docs.yml
@@ -14,6 +14,7 @@ on:
       - 'JuliaBUGS/docs/**'
       - 'JuliaBUGS/src/**'
       - '.github/workflows/Docs.yml'
+  workflow_dispatch:
 
 concurrency:
   # Skip intermediate builds: always.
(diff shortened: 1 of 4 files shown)


shravanngoswamii:
@sunxd3 @yebai Any idea about how can we avoid this!

shravanngoswamii:
> @sunxd3 @yebai Any idea about how can we avoid these failures!

Issue #27, opened 2022-10-31
Compile BUGS models to a `Distributions.jl` Distribution type


yebai:
Another very useful target for BUGS models is a `Distributions.Distribution` object. This would allow us to use these models as a new Distribution in other PPLs including Turing. 

Here are two concrete use cases: 

1. use BUGS language to specify a complicated hierarchical prior, then the DAG would allow us to perform re-parameterisation optimisation etc. See e.g. [here](https://julialang.slack.com/archives/CCYDC34A0/p1667205764190109?thread_ts=1667203312.461819&cid=CCYDC34A0) 
2. use BUGS language to specify an HMM/SSM model, then DAG would allow us to perform SMC/PMCMC inference efficiently

yebai:
Now that https://github.com/JuliaStats/Distributions.jl/pull/1803 is merged, can we give this another try? This fits nicely into the `to_distribution(bugs_model)` syntax.

seabbs:
This would be really great I think for supporting composing models from JuliaBugs with Turing.

PR #379, opened 2025-08-15
Add Distributions.jl target

sunxd3:
Some initial work for https://github.com/TuringLang/JuliaBUGS.jl/issues/27


sunxd3:
No worries! It's a draft anyway, totally okay to mess up, sometimes just have to poke the dragon...

Issue #400, opened 2025-10-07
example  6 does not compile


mhinsch:
Pasting the code of example 6 (magnesium) into a file `example_6.jl` and then running the following:

```Julia

using JuliaBUGS

struct Example{DNT <: NamedTuple, INT <: NamedTuple, INT2 <: NamedTuple, RNT}
           name::String
           model_def::Expr
           original_syntax_program::String
           data::DNT
           inits::INT
           inits_alternative::INT2
           reference_results::RNT
end

include("example_6.jl")

compile(model_def, data, inits)
```

produces an error:

```
ERROR: Function 'truncated' is not allowed in @bugs at #= line 0 =#. To use custom functions, declare them with @bugs_primitive macro.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] validate_bugs_expression(expr::Expr, line_num::LineNumberNode)
   @ JuliaBUGS ~/.julia/packages/JuliaBUGS/jScHL/src/JuliaBUGS.jl:210
 [3] validate_bugs_expression(expr::Expr, line_num::LineNumberNode)
   @ JuliaBUGS ~/.julia/packages/JuliaBUGS/jScHL/src/JuliaBUGS.jl:217
 [4] validate_bugs_expression(expr::Expr, line_num::LineNumberNode)
   @ JuliaBUGS ~/.julia/packages/JuliaBUGS/jScHL/src/JuliaBUGS.jl:228
 [5] compile(model_def::Expr, data::@NamedTuple{…}, initial_params::@NamedTuple{…}; skip_validation::Bool, eval_module::Module)
   @ JuliaBUGS ~/.julia/packages/JuliaBUGS/jScHL/src/JuliaBUGS.jl:260
 [6] compile(model_def::Expr, data::@NamedTuple{…}, initial_params::@NamedTuple{…})
   @ JuliaBUGS ~/.julia/packages/JuliaBUGS/jScHL/src/JuliaBUGS.jl:250
 [7] top-level scope
   @ REPL[9]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

sunxd3:
thanks @mhinsch , I'll fix this. In the meantime, could you try `@bugs_primitive truncated` and see it fixed the problem?

mhinsch:
> @bugs_primitive truncated

It does indeed (after loading Distributions and using `JuliaBUGS.@bugs_primitive truncated`).

PR #401, opened 2025-10-09
Add `truncated` and `censored` to the whitelist of functions

sunxd3:
fix https://github.com/TuringLang/JuliaBUGS.jl/issues/400


yebai:
Do Distributions.jl's `censored` and `truncated` share the same semantics with BUGS?

PR #402, opened 2025-10-09
DoodleBUGS: Now works in WebKit broswers (Safari, all iPad/iOS browsers)

### LINKED ISSUE #350: DoodleBUGS: Make it work for iPad
I don't know why Canvas isn't showing anything in iPads, let's fix this too!

shravanngoswamii:
Since last two days, I was doing trial and error on different parts of the codebase. Finally found the problem — `cytoscape-grid-guide` and `cytoscape-context-menus` were blocking touch events on WebKit browsers. Also, crypto.randomUUID() was not working properly in Safari and was causing node/edge creation to fail.

Now everything is fixed and working fine on all WebKit browsers including Safari and iPad.

I have also added a Debug Panel (can toggle it from `View` dropdown in navbar) for my personal usage — it’s useful for testing in iPad without needing a Mac.

Tested properly on iPad 11th Gen. @yebai @sunxd3 I’m on Windows, so if possible, please test this on Safari (Mac).

Closes #350

### DIFF
--- a/DoodleBUGS/src/components/common/DebugPanel.vue
+++ b/DoodleBUGS/src/components/common/DebugPanel.vue
@@ -1,12 +1,22 @@
-<!-- This file is AI generated using Claude Sonnet 4.5 -->
-<!-- TODO(shravanngoswamii): Review this file properly -->
 <script setup lang="ts">
-import { ref, onMounted, onUnmounted } from 'vue';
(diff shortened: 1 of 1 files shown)


sunxd3 (approved):
I think it's good to merge.

It's be good to note what functionalities are removed: grid snapping and right-click menu. (Is this correct?)

PR #418, opened 2025-12-07
DoodleBUGS: Build as Library for Integration in Any Web App

shravanngoswamii:
- DoodleBUGS can now be built as a distributable library, producing doodlebugs.js, doodlebugs.css, and doodlebugs.umd.cjs.
- The library output is suitable for direct use in other web apps via <script> and <link> tags, or as a custom element (<doodle-bugs>) in any frontend project.

https://turinglang.github.io/JuliaBUGS.jl/DoodleBUGS/pr-previews/418

https://turinglang.org/JuliaBUGS.jl/DoodleBUGS/pr-previews/418/DoodleWidget/

### DIFF
--- a/DoodleBUGS/docs/DoodleWidget/index.html
+++ b/DoodleBUGS/docs/DoodleWidget/index.html
@@ -4,11 +4,33 @@
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta name="description" content="DoodleWidget - Embeddable Bayesian graphical model editor component for JuliaBUGS">
-    <title>DoodleWidget - Interactive Demo</title>
+    
+    <title>DoodleWidget - Interactive Graphical Modeling for JuliaBUGS</title>
+    <meta name="title" content="DoodleWidget - Interactive Graphical Modeling for JuliaBUGS">
+    <meta name="description" content="Embeddable Bayesian graphical model editor for JuliaBUGS. Build, visualize, and export probabilistic models with an intuitive drag-and-drop interface.">
+    <meta name="keywords" content="Bayesian, graphical models, JuliaBUGS, BUGS, probabilistic programming, statistical modeling, DoodleBUGS, Turing, Julia, hierarchical models">
+    <meta name="author" content="Shravan Goswami, Hong Ge, Xianda Sun">
+    <meta name="robots" content="index, follow">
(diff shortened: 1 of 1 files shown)


shravanngoswamii:
Try DoodleWidget here: https://turinglang.org/JuliaBUGS.jl/DoodleBUGS/pr-previews/418/DoodleWidget/
Now we can integrate DoodleBUGS in any web app (Documenter.jl docs, Quarto documents or websites, etc...).

A minor change in bottom toolbar of main app too: https://turinglang.github.io/JuliaBUGS.jl/DoodleBUGS/pr-previews/418

Few new features:
- Add two new model examples (pumps and seeds), can be loaded from left sidebar.
- Adding new model examples is now much easier than before:
  - Just draw the model in DoodleBUGS, export it as a JSON and store that in `DoodleBUGS\public\examples\{model-name}\model.json` and add a entry of that model in `DoodleBUGS\src\config\examples.ts` with model name (same as directory name in examples) as id, name of model and url (raw github url).
  - Now this model can be used as default models in DoodleWidget with just id name.

DoodleWidget can be integrated like this (this will work after this PR is merged):
Add these in `<head>...</head>`:
```html
<link rel="stylesheet" href="https://turinglang.org/JuliaBUGS.jl/DoodleBUGS/lib/doodlebugs.css">
<script type="module" src="https://turinglang.org/JuliaBUGS.jl/DoodleBUGS/lib/doodlebugs.js"></script>
```

and use it anywhere on page like this:

```html
<doodle-bugs model="id | url (optional, if left empty then a blank canvas would appear)" height="optional" width="optional"></doodle-bugs>
```

shravanngoswamii:
> I don't have a lot of code nit to pick other than there is a bit of dup

I know there are few features which I duplicated, mostly I avoided duplicating code, most features are reused from main app, but there are few which were just creating issues when reused and those need proper lookup! Let's do that some other day!

> You could merge if you think it's ready.

I am still playing around few things, will merge soon!

yebai:
>>One thing I notice is that once the edit mode is enabled for one figure, all figures enter edit mode.

>That is actually a feature! So we don't have to click edit button again and again, toolbar and sidebars show content of the graph which is in the current view and if two graphs are in the same view then whichever is actively being interacted with

I was a bit confused by this as well — perhaps it wouldn’t be such a bad idea to require clicking the Edit button for each graph individually.

shravanngoswamii:
All looks good to me, let me know if you have any other suggestions! Feel free to merge!

Issue #327, opened 2025-06-12
Separating stochastic interpolation of missing data from the model


sunxd3:
Currently, missing data and "forward-sample"d stochastic variables will be sampled by the sampler of choice.

Ideally, they can be taken out of the regular sampling process and compute at once using the posterior samples.

This would need to modify `BUGSModel` to keep track of the identity of these variables:
At the moment, a node can be one of three types, `deterministic`, `observation`, or `model parameter`. We can create a `VariableType` enum and add the `generated_quantities` type, for instance.

PR #443, opened 2026-03-28
add VariableType enum

hardikxk:
A start for #327 
Only added the enum for identifying nodes in this commit. Next we can use the implementation so the samplers can optionally skip these nodes during MCMC for better performance.


sunxd3:
Hardik, thanks for the PR. This is genuinely useful and definitely on the right track.

The shape of the code changes looks right: classifying the variables, taking them out of log-density evaluation, updating `dimension`, etc. all look right. A minor thing to watch out for is not having this classification in the hot path.

The second half would be to make sure the generated quantities are sampled correctly and efficiently when assembling MCMC chains. There is also a user-communication problem — it's better if we have a function that samples these variables. Right now `gen_chains` is not built for this yet. The challenge there is how to not evaluate the whole graph.

I think delivering the whole feature will touch almost all the files, so let's do it in a step-by-step fashion. For instance, we can deal with the classification first, then move on to pruning them from inference, then do the forward sampling. The source-gen code will be touched at the end.

sunxd3:
Don't hesitate asking me to elaborate

hardikxk:
@sunxd3 I wanted to verify since the classification work is mostly done, for actually evaluating the GQs, I was thinking of making a new compile function alongside the existing log-density function where we could only compile `postprocess_stochastic` variables (and their ancestors nodes and all) and rest would end up as `nothing` for efficiency. I wanted to ask you how to go about this and whether my assumptions of this being a faster approach are valid. 
Also BTW, 
>There is also a user-communication problem — it's better if we have a function that samples these variables

Im not quite sure about the public API and what all should be exposed and what should not be

sunxd3:
Hardik, sorry for the late reply. 

On your two questions:
* There are three evaluation mode: `UseGraph`, `UseGeneratedLogDensityFunction`, `UseAutoMarginalization`. Each would need they own treatment. `UseGraph` is the simplest in concept (which I see you have pushed a commit for; while the idea is correct, deterministic nodes would need to be updated too given the new samples of the generated quantities). For `UseGeneratedLogDensityFunction`, indeed a generated function is the right way to go. `UseAutoMarginalization` should be similar to `UseGraph`, but need to take into considerations of the discrete variables being marginalized out.
    * The three would need to be consistent
* For public API, I think designing this would take some extra work, we can definitely skip it and do it in future PRs

Like I said in the previous comment, this PR is bound to be a bigger one, to review it properly, it'll be better to break it down. Thanks for the effort till now, happy to elaborate on any of this.

hardikxk:
Thanks for the review! I have updated `UseGraph` and also added the gen function for `UseGeneratedLogDensityFunction`.

hardikxk:
@sunxd3  I had some questions about `UseAutoMarginalization` so as of now after `evaluate!!` function call the marginalized variables are just stale and their values are not used anywhere (or produces wrong results if used) so to handle this is it better to create a function to loop through the discrete variables and then find the conditional posterior which then gets used in sampling? I was not quite sure about this approach so wanted to ask before any review.

sunxd3:
Thanks, Hardik. Your instinct in that last question is on point. To make things clear:

* For discrete latents with observed descendants (the marginalized ones): these are summed out of the HMC target, so they're not sampled during inference and aren't part of the parameter vector. They don't need to be in the chain.
* For generated quantities with no marginalized-discrete ancestor: forward-sample from the continuous parameters; this works for discrete and continuous GQs alike, and your current `postprocess` handles it.
* The tricky one is a GQ (continuous or discrete) that depends, directly or indirectly, on a marginalized discrete latent $z$. To forward-sample that GQ we need a value for $z$, and since $z$ has observed descendants it must come from its conditional posterior $p(z \mid \theta_c, y)$ — given the data and the current continuous draw. So per retained HMC sample $\theta_c^{(s)}$:
    * re-run the same marginalization at $\theta_c^{(s)}$ to build the per-node weights, sample the latent from those weights instead of summing them away (jointly if there are several: draw one, fix it, draw the next conditioned on it), which gives one $z^{(s)}$ per continuous draw, so $(\theta_c^{(s)}, z^{(s)})$ is a draw from the joint $p(\theta_c, z \mid y)$
    * forward-sample the GQ from the recovered $z$.

An alternative approach to the third case is error or warning and postpone it to a later PR.

hardikxk:
Thanks a lot for the guidance. 

I went ahead and tried doing the marginalization sampler which loops through the marginalization order and for each marginalized discrete latent and uses `_marginalize_recursive()` and then calculating the posteriors in the end which would be used in sampling.

If you can review the code and see if its something we can proceed with then I can add proper tests for this or else I could do a revert on this and just throw an error (or warning - but I guess an error would be better to prevent wrong computations) for now.

sunxd3:
Thanks, Hardik. Yeah, the code looks correct, good work!

Sorry for the delay again. Once you add tests, I'll do a proper review of the PR.

hardikxk:
Thanks you so much! I have added the tests for auto marginalization as well. I guess its ready for a full review now so.

sunxd3:
@hardikxk could you break the PR down into smaller PRs so we could review it properly? right now it's changing too many pieces at once and I don't think I can meaningfully understand the implications of the changes made.

hardikxk:
Sure I can do that. Hopefully that can also help identify why the 2 CI tests are failing which locally the individual tests passed and I wasn't quite able to figure out why the tests were failing.

I break it apart into 3 PRs or 4 for more simplicity.

hardikxk:
@sunxd3 ive added the 4 PRs that have picked the commits off of the branch / this PR. Doing a review there would be easier now and then after any individual changes I'll just merge it all back here on this PR 

Or I guess it would be better to review each PR and merge them one after another to make sure nothing breaks and we dont have any regression issues

sunxd3:
see https://github.com/TuringLang/JuliaBUGS.jl/pull/495#issuecomment-4787338072m thanks!

PR #451, opened 2026-05-03
fix: prevent DoodleBUGS global CSS from polluting host applications

shravanngoswamii:
Removes body and #app global styles from `global.css` and moves them to `App.vue` so they do not leak into third-party host applications like Documenter.jl or other consumers of DoodleWidget.


shravanngoswamii:
This PR is related to https://github.com/TuringLang/JuliaBUGS.jl/tree/Redesign-BUGSExamples -- both still work in progress

shravanngoswamii:
moved to https://github.com/mcmcjs/mcmcjs

PR #452, opened 2026-05-15
DoodleBUGS: emit per-language code updates from the widget

shravanngoswamii:
Replaces the single `code-update` event with two typed events: `bugs-code-update` and `stan-code-update`. Both fire whenever the model changes, independent of which language the user has visible in the code preview panel, so host applications that need one specific language no longer have to rely on the user picking it in the UI.

`useWidgetEmitter` now takes two refs (`generatedBugsCode`, `generatedStanCode`) and emits the two new events. `DoodleWidget.vue` updates its `defineEmits` and passes both refs into the emitter. The widget events table in `README.md` is updated to document all emitted events.

The previous `code-update` event is removed.

### DIFF
--- a/DoodleBUGS/src/composables/useWidgetEmitter.ts
+++ b/DoodleBUGS/src/composables/useWidgetEmitter.ts
@@ -7,13 +7,15 @@ import { useDataStore } from '../stores/dataStore'
 
 export interface WidgetEmitFn {
   (e: 'state-update', payload: string): void
-  (e: 'code-update', payload: string): void
+  (e: 'bugs-code-update', payload: string): void
+  (e: 'stan-code-update', payload: string): void
   (e: 'ready', payload: string): void
 }
 
 export function useWidgetEmitter(
   emit: WidgetEmitFn,
-  generatedCode: Ref<string>,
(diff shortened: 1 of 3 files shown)


DoodleBUGS/src/composables/useWidgetEmitter.ts
```diff
@@ -7,13 +7,15 @@ import { useDataStore } from '../stores/dataStore'
 
 export interface WidgetEmitFn {
   (e: 'state-update', payload: string): void
-  (e: 'code-update', payload: string): void
+  (e: 'bugs-code-update', payload: string): void
+  (e: 'stan-code-update', payload: string): void
```
sunxd3, line 11: README has `models-available`, just a heads-up

sunxd3 (approved):
LGTM

DoodleBUGS/src/composables/useWidgetEmitter.ts
shravanngoswamii: Yep, I did add that in some old PR but forgot to document it back then. It's useful if we make some self-hosted Pluto-based or some backend for running models.

PR #454, opened 2026-05-19
Use AbstractPPL prepared evaluator API for gradients

### LINKED ISSUE #449: Use native gradient API
Let's switch to native gradient API when https://github.com/TuringLang/AbstractPPL.jl/pull/155 is ready.

shravanngoswamii:
Switch BUGSModelWithGradient to prepare gradients through AbstractPPL's native prepared evaluator API.

The LogDensityProblems gradient method copies the gradient out of AbstractPPL's cache-backed bang API before returning it.

Closes #449


shravanngoswamii:
@yebai I am happy to open this PR for reviews. I need help in deciding whether we should keep DI as JuliaBUGS deps or not. If we keep `using DI` statements in JuliaBUGS, then users can just use ForwardDiff or ReverseDiff directly. Or else we can update docs to say users must load DI themselves before using ForwardDiff/ReverseDiff through AbstractPPL.

yebai:
> update docs to say users must load DI themselves 

Let's take the above approach.

JuliaBUGS/Project.toml
```diff
@@ -47,13 +46,12 @@ JuliaBUGSMCMCChainsExt = "MCMCChains"
 [compat]
 ADTypes = "1.14.0"
```
sunxd3:
```suggestion
ADTypes = "1.15.0"
```
https://github.com/SciML/ADTypes.jl/releases

JuliaBUGS/src/model/logdensityproblems.jl
```diff
@@ -110,21 +115,39 @@ function BUGSModelWithGradient(model::BUGSModel, adtype::ADTypes.AbstractADType)
...
+    return _is_mooncake(adtype) || adtype isa ADTypes.AutoEnzyme
+end
 
 function _check_ad_compatibility(model::BUGSModel, adtype::ADTypes.AbstractADType)
     if model.evaluation_mode isa UseGeneratedLogDensityFunction &&
         !_supports_mutation(adtype)
         @warn "AD backend $(typeof(adtype)) does not support mutation required by " *
             "UseGeneratedLogDensityFunction mode. Switching to UseGraph mode." maxlog = 1
         return set_evaluation_mode(model, UseGraph())
+    elseif model.evaluation_mode isa UseGraph && _is_mooncake(adtype)
```
sunxd3: also check `UseAutoMarginalization`

sunxd3 (commented):
thanks Shravan, two small issues

JuliaBUGS/Project.toml
shravanngoswamii: I was just committing this and you reviewed :smile: , 1.22 is available, should we bump to latest?

sunxd3: That would be fine for me

shravanngoswamii: I have tested with 1.18 and everything is fine. Let's not force users to latest...

sunxd3:
thanks! just making sure, if you haven't, could you run the examples in the doc to make sure it does run? (or maybe turn it into an actual doc example?)

sunxd3:
@shravanngoswamii it seems that Mooncake's issue with `UseGraph` might be this function
https://github.com/TuringLang/JuliaBUGS.jl/blob/c2860a574968dc17bcca7747b1e2ae7e07104a3e/JuliaBUGS/src/model/utils.jl#L131-L179

could you try and see if Mooncake works with 

```julia
function smart_copy_evaluation_env(env::NamedTuple, mutable_syms::Set{Symbol})
    new_env = env
    for k in keys(env)
        if k in mutable_syms
            new_env = BangBang.setindex!!(new_env, deepcopy(env[k]), k)
        end
    end
    return new_env
end
```

JuliaBUGS/src/model/logdensityproblems.jl
```diff
@@ -110,21 +119,47 @@ function BUGSModelWithGradient(model::BUGSModel, adtype::ADTypes.AbstractADType)
     model = _check_ad_compatibility(model, adtype)
 
     x = getparams(model)
-    prep = DI.prepare_gradient(_logdensity_for_gradient, adtype, x, DI.Constant(model))
+    prep = Base.invokelatest(_prepare_logdensity_gradient, adtype, model, x)
```
yebai: It would be helpful to understand why this is needed and, if possible, remove it.

shravanngoswamii: Thanks, I checked this again. I removed the broad `Base.invokelatest` around `_prepare_logdensity_gradient`.

The only remaining `Base.invokelatest` is inside the generated log-density branch, around `AbstractPPL.prepare(...)`. That seems needed because the generated log-density function is created dynamically with `Core.eval`, and `AbstractPPL.prepare` calls the target immediately while preparing AD.

I avoided putting `Base.invokelatest` inside the function being differentiated, because Mooncake cannot differentiate through that.

shravanngoswamii:
> could you try and see if Mooncake works with

Yes, it did work! Thanks @sunxd3!

shravanngoswamii:
I re-tested after the `smart_copy_evaluation_env` change. `AutoMooncake()` reverse mode now works with `UseGraph()` and also with `UseAutoMarginalization()` in the added tests. `AutoMooncakeForward()` still needs generated log-density mode.

Do you think we should make `AutoMooncake()` the first recommendation in the AD docs now, or keep ReverseDiff first for this PR and do that wording change separately after more model coverage?

yebai:
>  we should make AutoMooncake() the first recommendation in the AD docs now,

Good idea to default to Mooncake.

shravanngoswamii:
> if you haven't, could you run the examples in the doc to make sure it does run? (or maybe turn it into an actual doc example?)

Yep, I did ran locally and tested it. Works fine. Turned the doc to be executable example now!

shravanngoswamii:
@yebai @sunxd3 This is ready for final review!

JuliaBUGS/src/model/bugsmodel.jl
sunxd3: need to use `invokelatest` here too.

JuliaBUGS/Project.toml
sunxd3: If I am not mistaken, `LogDensityProblemsAD` can go? We probably should've removed it in a past pr.

sunxd3 (approved):
Thanks a lot @shravanngoswamii, looks very good.

https://github.com/TuringLang/JuliaBUGS.jl/tree/main/JuliaBUGS/examples would need updates. We can do this in another PR and maybe add these examples to documentation?

sunxd3:
Okay to merge on my side

shravanngoswamii:
> https://github.com/TuringLang/JuliaBUGS.jl/tree/main/JuliaBUGS/examples would need updates. We can do this in another PR and maybe add these examples to documentation?

Lets improve them in some later PR but let me update them in this PR to use Mooncake for now.

sunxd3:
(I didn't roll them into documentation because, at least, bnn.jl take quite a long time. But maybe there are some easy optimization I could've done.)

shravanngoswamii:
> (I didn't roll them into documentation because, at least, bnn.jl take quite a long time. But maybe there are some easy optimization I could've done.)

Aha, it does take tame...

shravanngoswamii:
@sunxd3 Can you give it a final look and let me know if we are ready to merge this. Feel free to be nitpicky. Thanks!

JuliaBUGS/Project.toml
shravanngoswamii: removed!

JuliaBUGS/src/model/bugsmodel.jl
shravanngoswamii: done, thanks!

shravanngoswamii:
@yebai Ready for final look!

sunxd3 (approved):

yebai (approved):

PR #459, opened 2026-05-21
Add to_distribution: expose BUGSModel as a Distributions.Distribution (closes #27)

### LINKED ISSUE #27: Compile BUGS models to a `Distributions.jl` Distribution type

yebai:
Closes #27.

Adds `to_distribution(model::BUGSModel)` returning a `BUGSModelDistribution <: Distribution{NamedTupleVariate{names}, S}`, where `names` are the unique parameter symbols in graph evaluation order. `rand` does ancestral sampling and returns a `NamedTuple` (preserving observed slots in partially-observed arrays); `logpdf` overlays the supplied `NamedTuple` into the evaluation environment and returns the joint log density in the original (constrained) space, regardless of `model.transformed`. This gives BUGS models a standard `Distributions.jl` interface, covering the hierarchical-prior and HMM/SSM use cases the issue calls out.

`Distributions.jl` has no array fallback for `Distribution{NamedTupleVariate}`, so the scalar `rand` alone would let `rand(rng, d, n)` recurse into `Random`'s array path and blow the stack. The PR therefore also defines `Base.rand(rng, d, ::Dims)` and `Distributions._rand!`, covering `rand(d, n)`, `rand(rng, d, n)`, and `rand!(rng, d, X)`.

Also adds a documentation page ("Model as a Distribution") describing the interface, its semantics (constrained space, full joint density, observed data baked into the model), and its limitations.

## Future work

Support for integrating these distributions into DynamicPPL (e.g. using a BUGS model as a `~`-distributed sub-model in a Turing model) is intentionally **not** part of this PR and will be addressed in a future PR. This PR is self-contained: it only adds the `Distributions.jl` wrapper itself.


JuliaBUGS/src/JuliaBUGS.jl
```diff
@@ -41,6 +41,18 @@ include("model/Model.jl")
 using .Model
 using .Model: AbstractBUGSModel, BUGSModel
 
+# Re-export the user-facing API of the Model submodule.
+export parameters,
```
yebai: These probably should not be exported.

PR #489, opened 2026-06-17
1. Variable type enum setup

hardikxk:
#443 
1
- Added a `VariableType` enum - Deterministic, Observation, ModelParameter, GeneratedQuantity to classify each node in the model graph. 
- Splits stochastic parameters into `mcmc_parameters` and `postprocess_stochastic`

### DIFF
--- a/JuliaBUGS/src/model/Model.jl
+++ b/JuliaBUGS/src/model/Model.jl
@@ -27,10 +27,10 @@ include("to_distribution.jl")
 # Public user-facing API
 export parameters, variables, initialize!, getparams, settrans, to_distribution
 export set_evaluation_mode, set_observed_values!
-export mcmc_parameters, postprocess_variables
+export model_parameters, generated_quantities, variable_type
 
 # Variable classification
-export VariableType, Deterministic, Observation, ModelParameter, GeneratedQuantity
+export VariableType, Observation, ModelParameter, TransformedParameter, GeneratedQuantity
 
 # Evaluation mode types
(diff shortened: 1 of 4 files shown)


sunxd3:
Thanks a lot for this, Hardik.

Would `@enum VariableType Observation ModelParameter TransformedParameter GeneratedQuantity` make more sense?
* `TransformedParameter` are deterministic nodes that needs to be computed during each iteration of sampling
* `GeneratedQuantity` contain both stochastic and deterministic variables that can be computed after getting samples

And the field can be:
* `variable_types::Vector{VariableType}` (it needs to be consistent to `sorted_nodes`)
* `model_parameters` (this is what you called `mcmc_parameters` right now, renaming so it's not algorithm bounded)
* `generated_quantities` (instead of "postprocess")

For this PR, we can focus on the categorization and the accessor functions:
* `model_parameters`
* `generated_quantities`
* `variable_type(model, vn)`

Think about testing and what invariant should hold for these categories.

hardikxk:
Sure, the renames would help kinda generalize things more I guess. I just did them the way to keep it simple and first hand.

sunxd3:
looks good, could you merge to https://github.com/TuringLang/JuliaBUGS.jl/tree/variabletype_enum instead? we will do a non-sqaush merge into main in the end. This is just so that we can temporarily break the branch if needed.

hardikxk:
@sunxd3 should be good to go

sunxd3:
Thanks!

PR #490, opened 2026-06-17
2. Route model_parameters

hardikxk:
#443 
2

- Use `model_parameters` instead of all stochastic variables. Generated quantities are now excluded from the sampler's parameter vector
- added `_active_parameters` in `bugsmodel.jl` to and check and filter models based on the parameters passed.
- changed `GraphEvaluationData` to cleanly separate `model_parameters `from `generated_quantities`  
- updated `evaluate_with_env!!` to skip generated quantities from `logprior` calculation by default

the follow up to this will have
- forward pass implementation in source_gen.jl to be able to forward sample generated quantities 
- updating the to_distribution.jl to calculate the full joint density by finding the prior for generated quantities.

### DIFF
--- a/JuliaBUGS/src/compiler_pass.jl
+++ b/JuliaBUGS/src/compiler_pass.jl
@@ -839,7 +839,7 @@ function analyze_statement(pass::AddVertices, expr::Expr, loop_vars::NamedTuple)
     args, node_function_expr, node_function = pass.f_dict[expr]
 
     vn = if lhs isa Symbol
-        AbstractPPL.VarName{lhs}(AbstractPPL.Iden())
+        AbstractPPL.VarName{lhs}()
     else
         v, indices... = lhs
         AbstractPPL.VarName{v}(AbstractPPL.Index((indices...,), (;)))
--- a/JuliaBUGS/src/model/abstractppl.jl
+++ b/JuliaBUGS/src/model/abstractppl.jl
@@ -89,11 +89,11 @@ julia> # Check parameter lengths
        model.untransformed_param_length  # Original has 4 parameters
 4
 
-julia> model_cond.untransformed_param_length  # After conditioning x[1], x[2]
-2
+julia> model_cond.untransformed_param_length  # x[3], y are now Generated Quantities
+0
 
-julia> model_cond2.untransformed_param_length  # After conditioning all x[i]
-1
+julia> model_cond2.untransformed_param_length  # y is now a Generated Quantity
+0
 
 julia> # NamedTuple syntax
        model_cond3 = condition(model, (; y=10.0));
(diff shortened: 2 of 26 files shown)


hardikxk:
@sunxd3 this is ready for review

JuliaBUGS/src/model/bugsmodel.jl
```diff
@@ -506,19 +546,10 @@ params_vec = getparams(model, custom_env)
 params_dict = getparams(Dict, model, custom_env)
 ```
 """
+
+
```
sunxd3:
```suggestion
```

JuliaBUGS/src/compiler_pass.jl
```diff
@@ -839,7 +839,7 @@ function analyze_statement(pass::AddVertices, expr::Expr, loop_vars::NamedTuple)
     args, node_function_expr, node_function = pass.f_dict[expr]
 
     vn = if lhs isa Symbol
-        AbstractPPL.VarName{lhs}(AbstractPPL.Iden())
+        AbstractPPL.VarName{lhs}()
```
sunxd3, line 842: is there a reason for this change?

sunxd3 (changes requested):
Thanks!

With this slice, I think it's sufficient to take the generated quantities out of MCMC sampling for the `UseGraph` mode. I would suggest move forward-sampling, source-gen, and `to_distribution` changes to follow-ups.

Could you also write a bit more in the PR description on the scope, design, and items to be addressed in follow-up PRs?

JuliaBUGS/src/compiler_pass.jl
hardikxk: stumbled on this when running into some errors with v 0.15 of AbstractPPL. The change doesn't really make any difference since the Identity optic would just be initialized [automatically](https://github.com/TuringLang/AbstractPPL.jl/blob/0c49dbacffc3bdd338c6c99b040b20b0a61fdeac/src/varname/varname.jl#L11-L16) without having to explicitly pass it as in the original source and just looks cleaner. I can revert it to make it less abstract although there really isn't a point in passing the arg when its doing the thing anyway under the hood.

hardikxk:
Sure I'll restore back the source_gen changes and the rest and continue it in a follow up.

JuliaBUGS/src/model/abstractppl.jl
```diff
@@ -541,9 +542,13 @@ function _create_modified_model(
     new_evaluation_env::NamedTuple;
     base_model=nothing,
 )
-    # Create new graph evaluation data
-    new_graph_evaluation_data = GraphEvaluationData(new_graph)
-    new_parameters = new_graph_evaluation_data.sorted_parameters
+    # Create new graph evaluation data preserving the original GQ classification
+    # We must explicitly keep the original `generated_quantities` to avoid
```
sunxd3: unfinished sentence

JuliaBUGS/src/model/bugsmodel.jl
```diff
@@ -506,19 +546,9 @@ params_vec = getparams(model, custom_env)
 params_dict = getparams(Dict, model, custom_env)
 ```
 """
+
```
sunxd3: don't think we need this newline

JuliaBUGS/src/compiler_pass.jl
sunxd3: okay, it's fine to keep this change, then, thanks

JuliaBUGS/test/model/to_distribution.jl
```diff
@@ -82,42 +82,6 @@
         @test logpdf(d, nt) ≈ manual
     end
 
-    @testset "partially observed array" begin
```
sunxd3, line 85: any reason this test is removed?

sunxd3 (commented):
thanks! some final nits

JuliaBUGS/test/model/to_distribution.jl
hardikxk: Yes the test was failing after modification to `evaluate_with_env!!` and since `to_distribution` needs changes which are done in #495 where I am also adding these tests back in again as is.

I mean I it wasn't very good to just remove it even though the testset will be again added in the next PR but maybe if you want I'll just comment it out for this one?

sunxd3 (approved):
this is nice, thanks!

PR #491, opened 2026-06-17
3. Forward sample generated quantities

hardikxk:
#443 
3

- Added `_gen_postprocess_function_expr` generated quantities stochastic nodes now get forward-sampled from their distributions and deterministic nodes get recomputed


sunxd3:
https://github.com/TuringLang/JuliaBUGS.jl/pull/443#issuecomment-4787342050

PR #492, opened 2026-06-17
4. Margenalized latents

hardikxk:
#443 
4

- Added `evaluate_and_sample_with_marginalization_values` which reconstructs the discrete latent values by sampling from their conditional posteriors. 
- skips Stochastic nodes in `_marginalize_recursive` so models with stochastic nodes don't crash during auto marginalization.

- Also added necessary tests (might seem a little verbose and if wanted can refactor to make it less obtuse)


sunxd3:
https://github.com/TuringLang/JuliaBUGS.jl/pull/443#issuecomment-4787342050

PR #501, opened 2026-06-25
Variable-type classification and generated quantities

### LINKED ISSUE #327: Separating stochastic interpolation of missing data from the model

sunxd3:
fix https://github.com/TuringLang/JuliaBUGS.jl/issues/327

## Summary

Consolidates the `variabletype_enum` line of work into `main`. It introduces an explicit per-node **variable-type classification** and uses it to separate **generated quantities** — unobserved nodes (stochastic or deterministic) outside the log-density target dependency closure — from the MCMC target. Generated quantities are excluded from the log density in every evaluation mode and recovered afterwards by forward sampling, instead of being sampled during inference. Support is complete across all three evaluation modes, including discrete-latent recovery and chain reconstruction for auto-marginalized models.

## Classification

Every node is assigned a `VariableType`:

- `Observation` — stochastic node with observed data.
- `ModelParameter` — unobserved stochastic node that influences observations. These are what MCMC samples; they make up the parameter vector and `LogDensityProblems.dimension`.
- `TransformedParameter` — deterministic node needed to evaluate the target density.
- `GeneratedQuantity` — unobserved node (stochastic or deterministic) outside the target dependency closure; forward-sampled in post-processing.

Model parameters and generated quantities are disjoint, and no model parameter or observation ever has a generated quantity as an ancestor.

## Behavior

A generated quantity is excluded from the log density in **every** evaluation mode and recovered by forward sampling:

| | log density excludes GQ | forward-sample GQ |
|---|---|---|
| `UseGraph` | ✅ | ✅ |
| `UseGeneratedLogDensityFunction` | ✅ (generated source drops the GQ statements) | ✅ |
| `UseAutoMarginalization` | ✅ | ✅ (recovers marginalized discrete latents from `p(z\|θ,y)` first) |

The no-observation behavior is preserved: a model with no data keeps all unobserved stochastic nodes as model parameters (prior sampling); only terminal deterministic nodes that feed no stochastic factor remain generated quantities.

## Public API

New names, exported from the `JuliaBUGS.Model` submodule (like the existing `parameters`) — reachable as `JuliaBUGS.model_parameters`, etc., not via `using JuliaBUGS`:

- `model_parameters(model)`, `generated_quantities(model)`, `variable_type(model, vn)`.
- `VariableType` and its instances `Observation`, `ModelParameter`, `TransformedParameter`, `GeneratedQuantity`.

Plus `JuliaBUGS.Model.forward_sample_generated_quantities!!(rng, model[, env])` (unexported) — applied automatically by `gen_chains`, so reported generated quantities are genuine posterior(-predictive) draws.

## ⚠️ Breaking changes

- Unobserved stochastic nodes with **no observed descendants** (e.g. posterior-predictive draws, priors not used by any likelihood) are now **generated quantities**: excluded from `model_parameters`, the parameter vector, and `LogDensityProblems.dimension`, and forward-sampled in post-processing rather than sampled by MCMC/Gibbs. (The resulting joint is unchanged — forward-sampling a node from its conditional given its parents is identical to sampling it as a free parameter.)
- `parameters(model)` (all unobserved stochastic nodes) and `model_parameters(model)` (the MCMC target) now differ; `dimension` follows `model_parameters`. Code that assumed `length(parameters(model)) == dimension` should switch to `model_parameters` / `dimension`.
- `Gibbs` sampler maps that reference a now-reclassified variable will error, since it is no longer in `model_parameters`.

## What's included

| PR | Change |
|----|--------|
| #489 | Variable-type enum scaffolding (`Observation`, `ModelParameter`, `TransformedParameter`, `GeneratedQuantity`) |
| #490 | Route `model_parameters` through the classification |
| #496 | Unify the generated-quantity partition with a per-node mask |
| #497 | Forward-sample generated quantities (`forward_sample_generated_quantities!!`) |
| #498 | Support generated quantities in the generated log-density function |
| #499 | Generated quantities under auto-marginalization (log density + case-3 latent recovery) |
| #500 | Document generated quantities |
| #504 | End-to-end `gen_chains` support for auto-marginalized models (continuous-only environment reconstruction + recovered finite discrete latents) |
| #505 | Finalize cleanup: reclassify ancestors after conditioning, rename `gq_override → generated_quantities`, cache `continuous_model_parameters` on the marginalization cache |
| #507 | Fix the GQ target closure: reclassify prior-only deterministic ancestors as transformed parameters, derive the cached GQ classification by default in source generation, and align auto-marginalized `ParamsWithStats` with the continuous flat parameter vector |
| #508 | Shrink generated-source guards by selecting the smaller kept/dropped iteration set, and skip deterministic-GQ branches when recomputation is cheaper than guarding |

Plus two direct commits: dropping `invokelatest` from marginalized evaluation, and a `gibbs.jl` test update reflecting the reclassification.

## Tests

Each PR added focused tests; together they cover the classification, the per-node mask, generated-vs-graph log-density agreement, an invariant battery (mode-invariance, partition, GQ-independence, dimension), marginalized-target invariance against a forward-algorithm reference, backward-sampling correctness against an analytic HMM posterior (empirical distribution of the recovered discrete latents), and auto-marginalized `gen_chains` reconstruction from continuous-only samples. CI is green.

### DIFF
--- a/JuliaBUGS/History.md
+++ b/JuliaBUGS/History.md
@@ -1,5 +1,19 @@
 # JuliaBUGS Changelog
 
+## 0.15.0
+
+### Highlights
+
+- **Generated quantities (#501).** Every node now carries an explicit `VariableType`: `Observation`, `ModelParameter`, `TransformedParameter`, or `GeneratedQuantity`. A *generated quantity* is an unobserved node (stochastic or deterministic) with no observed descendants, so it lies outside the log-density target. Generated quantities are excluded from the log density in all three evaluation modes (`UseGraph`, `UseGeneratedLogDensityFunction`, `UseAutoMarginalization`) and recovered after sampling by forward simulation instead of being sampled by MCMC. Under auto-marginalization, a generated quantity that depends on a marginalized discrete latent first recovers that latent from its conditional posterior `p(z | θ, y)`. `gen_chains` applies this automatically, so reported generated quantities are genuine posterior(-predictive) draws.
+
+- **New API:** `model_parameters(model)`, `generated_quantities(model)`, `variable_type(model, vn)`, the `VariableType` enum and its instances, and `Model.forward_sample_generated_quantities!!`.
+
+### Breaking Changes
+
+- Unobserved stochastic nodes with **no observed descendants** (e.g. posterior-predictive draws, or priors unused by any likelihood) are now generated quantities: excluded from `model_parameters`, the parameter vector, and `LogDensityProblems.dimension`, and forward-sampled in post-processing instead of sampled by MCMC/Gibbs. The joint distribution is unchanged.
+- `parameters(model)` (all unobserved stochastic nodes) and `model_parameters(model)` (the MCMC target) can now differ, and `dimension` follows `model_parameters`. Use `LogDensityProblems.dimension(model)` (or `length(model_parameters(model))`) where you previously relied on `length(parameters(model))` as the parameter-vector length.
+- `Gibbs` sampler maps that reference a reclassified variable now error, since it is no longer in `model_parameters`.
+
 ## 0.14.1
 
 ### Highlights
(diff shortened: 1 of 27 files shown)


sunxd3:
@hardikxk I just merged this PR, it won't be released immediately. If you want to, could you take a look and give some comments?

sunxd3:
@yebai @shravanngoswamii for visibility. 

The PR would change parameter counts for some models during inference, but if the Chain extensions work correctly, it should not break DoodleBUGS and other integrations.

PR #506, opened 2026-06-29
Fix generated-quantity target closure

sunxd3:
## Summary

- Reclassify prior-only deterministic ancestors of stochastic parameters as transformed parameters so generated source recomputes them.
- Make source generation derive the cached generated-quantity classification by default.
- Skip excluded generated quantities before graph/env logpdf work and align auto-marginalized ParamsWithStats with the continuous flat parameter vector.
- Update generated-quantity docs and add focused regressions.

## Why

PR #501 introduced a mismatch where prior-only deterministic ancestors could be classified as generated quantities and dropped from generated log-density source, leaving stale environment values in the target. The generated-quantity rule now follows the log-density target dependency closure instead.

### DIFF
--- a/JuliaBUGS/docs/src/inference/generated_quantities.md
+++ b/JuliaBUGS/docs/src/inference/generated_quantities.md
@@ -1,10 +1,11 @@
 # Generated Quantities
 
-A **generated quantity** is an unobserved node — stochastic or deterministic — that has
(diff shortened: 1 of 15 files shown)


sunxd3:
Superseded by #507 on branch fix-generated-quantity-target-closure.

PR #509, opened 2026-06-30
Make `@bugs` / `@bugs"..."` return a callable `BUGSModelDef` (#383)

### LINKED ISSUE #383: Unify model construction workflow
Currently, model definition in JuliaBUGS is inconsistent:

-   Models defined with `@bugs` or `@bugs_str` require an explicit call to `compile(...)`.
-   Models defined with `@model` are directly callable and return a `BUGSModel`.

This inconsistency can be confusing and complicates the implementation.

```julia
# @bugs macro
model1 = @bugs begin 
   x[1:2] ~ dmnorm(mu[:], sigma[:, :]) 
   x[3] ~ dnorm(0, 1) 
   y = x[1] + x[3] 
end

# @bugs_str macro
model2 = @bugs"""
  model {
     x[1:2] ~ dmnorm(mu[:], sigma[:, :]) 
     x[3] ~ dnorm(0, 1) 
     y = x[1] + x[3] 
  }
"""

# @model macro
@model function model3((;x), mu, sigma)
   x[1:2] ~ dmnorm(mu[:], sigma[:, :]) 
   x[3] ~ dnorm(0, 1) 
   y = x[1] + x[3] 
end
```

Usage:

```julia
# Requires compile(...)
compile(model1, (;mu=[0,0], sigma=[1 0; 0 1]))
compile(model2, (;mu=[0,0], sigma=[1 0; 0 1]))

# Directly callable
model3((;), [0,0], [1 0; 0 1])
```

Unify the API so that **all models—whether defined via `@bugs`, `@bugs_str`, or `@model`—return a callable function** of the form

```julia
(data::NamedTuple, other_args::NamedTuple) -> BUGSModel 
```

That is, let `@bugs` and `@bugs_str` return an anonymous function instead of a static object.

```julia
# unified behaviour
m1 = model1((;mu=[0,0], sigma=[1 0; 0 1])) # --> `compile(model1, (;mu=[0,0], sigma=[1 0; 0 1]))`
m2 = model2((;mu=[0,0], sigma=[1 0; 0 1])) # --> `compile(model2, (;mu=[0,0], sigma=[1 0; 0 1]))`
m3 = model3((;mu=[0,0], sigma=[1 0; 0 1])) # --> `compile(model3, (;mu=[0,0], sigma=[1 0; 0 1]), ; skip_validation=true)`

# The current `@model` generated constructor would still work. 
# But, this is not available for models defined via `@bugs` and `@bugs_str` 
m3b = model3((;), [0,0], [1 0; 0 1])

# all models can be initialised via
initialize!(m1::BUGSModel, initial_params::NamedTuple)
```

Then, we can remove `compile` from the export list and keep it only as an internal utility.

sunxd3:
Fixes https://github.com/TuringLang/JuliaBUGS.jl/issues/383

## What

`@bugs` and `@bugs"..."` now return a callable `BUGSModelDef` instead of a bare `Expr`. Calling the definition with a data `NamedTuple` constructs a `BUGSModel`:

```julia
seeds = @bugs begin
    # ...
end
model = seeds(data)   # was: compile(seeds, data)
```

This mirrors DynamicPPL — where the macro yields a callable model definition you invoke with data — and makes `compile` an implementation detail rather than a step users invoke directly.

## Changes

- `@bugs` / `@bugs_str` build a `BUGSModelDef` wrapping the model-definition `Expr`; the struct is callable: `model_def(data; kwargs...) -> BUGSModel`.
- `compile(::BUGSModelDef, ...)` unwraps to the existing `Expr` method, so existing `compile(@bugs(...), data)` call sites keep working. `compile` is **retained for backward compatibility** and is now mostly internal.
- The `@bugs` macros moved from the `Parser` submodule up to `JuliaBUGS`. `BUGSModelDef` is intentionally **not exported** (symmetric with `BUGSModel`).
- Pass-level tests use the internal `Parser.bugs_top` to obtain the `Expr` directly, instead of building-and-unwrapping `@bugs`.
- Docs updated to the callable idiom: each example's definition is named after its model (DynamicPPL-style — `seeds`, `normal_model`, ...), `model` is reserved for the constructed `BUGSModel`, and `compile` is framed as backward-compat/internal.

## Notes

- `BUGSModel.model_def` stays an `Expr`, so serialization and source generation are unaffected.


shravanngoswamii:
Ahh, just saw this -- Looks great to me, Thank you @sunxd3!

PR #523, opened 2026-07-31
let sampling return a vector of ParamsWithStats

shravanngoswamii:
`sample` now returns `ParamsWithStats` by default: one entry per draw, params keyed by `VarName` so array variables stay whole, stats as the sampler reported them. One vector per chain with the parallel samplers. `AbstractMCMC.from_samples(Chains, reshape(draws, :, 1))` and the same with `VNChain` convert them. Samplers JuliaBUGS has no method for keep AbstractMCMC's behaviour and hand back raw transitions.

`rand([rng,] model)` draws from the prior and returns the evaluation environment, an instance of `of(model)`. Observed variables keep their data values.

Samplers implement one method, `transition_params_and_stats(model, sampler, transition)`, and all output formats are built from it.

Other changes, breaking ones marked:

- sampling without a `chain_type` returns `Vector{ParamsWithStats}` instead of raw sampler transitions (breaking)
- callbacks report `VarName` keys in the model's space for every sampler; with HMC, MH and slice they used to give unnamed `θ[i]` in unconstrained space (breaking: `pws.params.mu` -> `pws.params[@varname(mu)]`)
- Gibbs and IndependentMH chains carry `lp`; `internals` was empty before (breaking: one extra column)
- a statistic a draw did not report is absent instead of padded (breaking: in `VNChain` it is `missing`, not `NaN`)
- `gen_chains` takes one `NamedTuple` per draw, not a names list plus a values table (breaking)
- MALA columns named after the model instead of `param_1`

The params container is `OrderedDict{VarName,Any}` for now, as agreed. `VarNamedTuple` is not reachable from here yet. Swapping it later is contained: the two hand-written `from_samples` methods can go once FlexiChains can read the draws directly.


shravanngoswamii:
I forgot that it was already discussed to stay consistent with DynamicPPL -- approach here is not consistent so let me make the changes before asking for review again.

shravanngoswamii:
#517 and TuringLang/AbstractPPL.jl#178, both closed. TuringLang/AbstractMCMC.jl#212 landed, so the wrapper is there. The container is not.

this PR keeps `OrderedDict{VarName,Any}` in the params slot. Now, FlexiChains cannot read it directly, so I wrote `from_samples` for `Chains` and `VNChain` by hand. Those can be deleted if the container becomes VarNamedTuple.

Should I wait for TuringLang/AbstractPPL.jl#178, or keep this container and swap later? Making `chain_type=Any` return `ParamsWithStats` and `rand` return an of-type object are both independent of that, so I can do those either way.

@sunxd3 Can you help me understand things here like in what direction do we want to go here?

JuliaBUGS/test/runtests.jl
```diff
@@ -92,6 +92,7 @@ const TEST_GROUPS = OrderedDict{String,Function}(
         include("model/to_distribution.jl")
     end,
     "callbacks" => () -> include("model/abstractmcmc.jl"),
+    "params_with_stats" => () -> include("model/params_with_stats.jl"),
```
sunxd3, line 95: **This group never runs in CI.** `.github/workflows/Tests.yml` enumerates `test_group` explicitly and `params_with_stats` is not in the list, so all 214 lines of `test/model/params_with_stats.jl` — the tests for this PR's headline feature — are skipped. The 41 green checks on this PR do not cover it.

```yaml
        test_group:
          ...
          - "callbacks"
          - "params_with_stats"   # <- add
          - "inference_mh"
```

Worth adding before merge; I'd want to see that job green given how much of the new behaviour (`chain_type=Any`, the `OpaqueSampler` fallback, auto-marginalization recovery, the array-aliasing check) lives only there.

---
_Generated by [Claude Code](https://claude.ai/code)_

JuliaBUGS/src/model/abstractmcmc.jl
```diff
@@ -1,71 +1,145 @@
...
+    AbstractMCMC.ParamsWithStats(model, sampler, transition, state; params, stats, extras)
 
-    return AbstractMCMC.ParamsWithStats(p, s, e)
-end
+Extract a callback's view of a draw: the sampled parameters keyed by `VarName`, and the
+statistics the sampler reported for that step.
 
+Generated quantities are *not* included. They are forward-sampled once at the end of
+sampling, so they only appear in the sampling output (`chain_type`), not here. Samplers that
+report no statistics of their own get the model's log density under `lp`.
```
sunxd3: Generated quantities are not the only difference between the callback view and the sampling output — marginalized discrete latents are too, and that one isn't documented anywhere.

`named_parameters` (line 56) iterates `sampled_parameters`, which under `UseAutoMarginalization` returns `marginalization_cache.continuous_model_parameters`. `gen_chains`/`reconstruct_chain_values` iterates `model_parameters(model)`, which still contains the discrete latents. So for a marginalized model a callback sees `mu` only, while the output has `mu` **and** `z` — and `z` is not a generated quantity, so neither this docstring nor the "What a callback sees" section of `output_formats.md` accounts for it.

`test/model/abstractmcmc.jl` asserts `!haskey(pws.params, @varname(z))` and `test/model/params_with_stats.jl` asserts `d.params[@varname(z)] in (1, 2)`, so both behaviours are pinned — they just disagree, deliberately or not. Either recover the latents in the callback too, or say so in both places.

---
_Generated by [Claude Code](https://claude.ai/code)_

JuliaBUGS/ext/JuliaBUGSMCMCChainsExt.jl
```diff
@@ -188,34 +189,115 @@ function JuliaBUGS.gen_chains(
...
+`MCMCChains.Chains` stores scalars only. Column names are the union of the keys across draws,
+in first-seen order, with numeric array statistics expanded to one column per element
+(`key[i,j]`). Draws that do not report a column, and statistics that are not real numbers, get
+`NaN`.
+"""
+function flatten_stats(stats)
+    specs = Any[]
+    seen = Set{Symbol}()
+    for draw in stats, (key, value) in pairs(draw)
+        key in seen && continue
```
sunxd3: `seen` is keyed on `key` alone, so the first draw that reports a key fixes its column layout for the whole run. The `_slice_flat_info_stat_specs` this replaces keyed on `(key, index)`, which was strictly more forgiving. Concretely:

- key first seen as a 2-element array, later reported with 3 elements → element 3 is silently dropped (no column exists for it).
- key first seen as a scalar, later as an array → `stat_value` takes the `spec.index === nothing` branch, `value isa Real` is false, and every array draw gets `NaN`.
- key first seen as an array, later as a scalar → the scalar draws get `NaN` in all the `key[i]` columns.

`RandPermGibbs` reports a fixed-length `num_proposals`, so nothing in-tree hits this today, but silently dropping data on a shape change is a bad failure mode for a function whose whole job is "take the union of the keys across draws". Restoring `(key, index)` dedup is a two-line change and keeps the old semantics.

Separately, a non-numeric statistic now gets a column that is `NaN` in every draw; the old code emitted no column for it at all. An all-`NaN` entry in `internals` is noise — consider skipping keys that are never real-valued.

---
_Generated by [Claude Code](https://claude.ai/code)_

JuliaBUGS/ext/JuliaBUGSMCMCChainsExt.jl
```diff
@@ -188,34 +189,115 @@ function JuliaBUGS.gen_chains(
...
+    vals = [
+        vcat(
+            collect(Iterators.flatten(values(draws[i, j].params))),
+            stats_values[(j - 1) * niters + i],
+        ) for i in 1:niters, j in 1:nchains
+    ]
+    ncols = length(param_symbols) + length(stats_names)
+    all(row -> length(row) == ncols, vals) ||
+        throw(ArgumentError("draws do not all carry the same variables"))
+
```
sunxd3: Column *names* come from `first(draws).params`, but each row's *values* come from `Iterators.flatten(values(draws[i, j].params))` — that draw's own dict order. The `ncols` guard below catches a differing number of variables but not a differing order, so two draws with the same keys in a different order produce silently mislabelled columns rather than the `ArgumentError` the guard implies.

Every in-tree producer builds the dict in a fixed order (`gen_chains` walks `param_vars` then `generated_vars`; `named_parameters` walks `sampled_parameters`), so this is a robustness gap rather than a live bug — but `from_samples` takes any `AbstractMatrix{<:BUGSParamsWithStats}`, including vectors a user assembled from a callback or `reduce(hcat, …)` across heterogeneous runs.

Looking up by key instead of relying on iteration order fixes it and makes the guard honest:

```julia
param_keys = collect(keys(first(draws).params))
...
vcat(collect(Iterators.flatten(draw.params[k] for k in param_keys)), stats_values[...])
```

with a `keys(draw.params) == param_keys` check (or a `KeyError` from the lookup) standing in for the length comparison.

---
_Generated by [Claude Code](https://claude.ai/code)_

JuliaBUGS/src/model/abstractmcmc.jl
```diff
@@ -106,22 +183,114 @@ function reconstruct_chain_values(rng::Random.AbstractRNG, model::BUGSModel, sam
...
-    param_samples = Vector{Vector{Float64}}()
-    for env in evaluation_envs
-        model_with_env = Accessors.@set model.evaluation_env = env
-        push!(param_samples, getparams(model_with_env))
+    gen_chains(
+        ::Type{<:AbstractVector{<:AbstractMCMC.ParamsWithStats}},
+        model::BUGSModel,
+        samples,
+        stats_names,
+        stats_values;
```
sunxd3: Docstring signature is stale — it still lists `stats_names, stats_values` while the method below takes a single `stats`. The FlexiChains and MCMCChains docstrings were both updated; this one was missed.

---
_Generated by [Claude Code](https://claude.ai/code)_

sunxd3 (commented):
Reviewed the full diff statically (no Julia in my environment, so I did not run anything).

The core refactor is a clear win: collapsing five sampler×output-format extensions into one `transition_params_and_stats` per sampler plus one `bundle_transitions` is the right factoring, the ambiguity-resolution extensions are correctly scoped, dropping `MCMCChains` from the `AdvancedHMC`/`AdvancedMH` extension triggers is a real fix, and `output_formats.md` is unusually good — the "Supporting a new sampler" and "What a callback sees" sections do the work most PPL docs skip. `History.md` is honest about every break.

Five things, one of which I'd want fixed before merge.

**Must fix — the new tests don't run.** `test/runtests.jl` registers a `params_with_stats` group but `.github/workflows/Tests.yml` enumerates `test_group` explicitly and doesn't list it. `test/model/params_with_stats.jl` is 214 lines covering `chain_type=Any`, the `OpaqueSampler` fallback, the multi-chain shape, per-draw array copying, and marginalization recovery — none of it executes. Inline on `runtests.jl`.

**Should fix**
- The callback and the sampling output disagree on marginalized discrete latents, and only the generated-quantities difference is documented. Inline on `abstractmcmc.jl`.
- `flatten_stats` dedups on `key` rather than `(key, index)`, narrower than the `_slice_flat_info_stat_specs` it replaces; a stat whose shape changes across draws is silently truncated to the first draw's layout. Inline on `JuliaBUGSMCMCChainsExt.jl`.
- `from_samples(Chains, …)` takes column names from the first draw but values from each draw's own dict iteration order; the `ncols` guard doesn't catch reordering. Inline.
- Stale docstring signature on the `ParamsWithStats` `gen_chains`. Inline.

**Nits, take or leave**
- `gen_chains`'s positional `stats` shares a name with the `stats` keyword `AbstractMCMC.mcmcsample` forwards to `bundle_samples` (its `SamplingStats`). Harmless now — it lands in `kwargs...` and is ignored — but if anyone later adds a `stats` keyword to `gen_chains` it will silently pick up the wrong object. `draw_stats` would remove the trap.
- "No statistics" now has to be spelled `fill(NamedTuple(), length(samples))`; the natural `[]` throws `DimensionMismatch` out of `map(stats, log_densities)` inside `stats_with_log_density`. The three test call sites had to be rewritten for this. Accepting an empty collection or `nothing` would be friendlier for the public `gen_chains`.
- `rand(model, 100)` doesn't work — it falls through to `Base.rand(X, dims...)` and dies in `Random.Sampler`. `to_distribution.jl:130` defines the `dims` form explicitly for exactly this reason; the new `Base.rand(::BUGSModel)` doesn't. Either add the method or say "one draw per call" in the docstring, since `model_as_distribution.md` now shows `rand` twice with two different return shapes.
- `from_samples` uses `MCMCChains.concretize`, which is unexported.
- `AbstractMCMC.ParamsWithStats` calls `require_transition_params_and_stats` unconditionally, so a callback constructed with `params=false, stats=false` still throws for an unknown sampler even though it asked for nothing.

Nothing here suggests the design is wrong — `transition_params_and_stats` as the single extension point is the correct shape, and keeping `OrderedDict{VarName,Any}` with two hand-written `from_samples` methods to delete later is a reasonable place to stop while AbstractPPL#178 is open.

---
_Generated by [Claude Code](https://claude.ai/code)_

JuliaBUGS/src/model/abstractmcmc.jl
```diff
@@ -106,22 +183,114 @@ function reconstruct_chain_values(rng::Random.AbstractRNG, model::BUGSModel, sam
...
+
+Unlike the chain formats, a plain vector carries no iteration indices, so `discard_initial`
+and `thinning` leave no trace in the result.
+"""
+function JuliaBUGS.gen_chains(
+    ::Type{<:AbstractVector{<:AbstractMCMC.ParamsWithStats}},
+    model::BUGSModel,
+    samples,
+    stats;
+    rng::Random.AbstractRNG=Random.default_rng(),
```
sunxd3: there are some annoyance and name clush with MCMCThreads apparently

---

context for agents:

>Make stochastic generated-quantity reconstruction reproducible and thread-safe: the positional sampling RNG is currently ignored during bundling, while a keyword RNG is shared across `MCMCThreads`; establish independent per-chain post-processing RNGs and test repeated seeded single- and multi-chain runs.

JuliaBUGS/test/runtests.jl
sunxd3, line 95: the test still need to be added to action to be ran (right now I don't think it is)

sunxd3 (commented):
look good in general

some small points under, and additionally:
1. MH returns `accepted`, we should return this too, this was not handled great before

JuliaBUGS/src/model/abstractmcmc.jl
shravanngoswamii: Went a different way than per-chain RNG plumbing: the reconstruction stream is seeded from the draws plus the chain number. Draws are already reproducible from the sampling seed, so reconstruction is too, and chains never share state.

sunxd3 (approved):
thanks @shravanngoswamii 

I did a commit on some nits, could you give it a look? if good, I think it's ready

shravanngoswamii:
@sunxd3 Thank you, looks good to me. Can we merge it now?

sunxd3:
yes!

PR #525, opened 2026-08-20
Fix T()/C() allowlist, dweib parameterization, and dt initialization

sunxd3:
Three small, independent bug fixes in BUGS primitives and the default function allowlist. Each was found while auditing how BUGS syntax is supported and is reproduced on `main` below.

## 1. `T(,)` / `C(,)` programs could not compile

```julia
compile(@bugs("model { x ~ dnorm(0, 1)T(0, ) }"), (;))
# ERROR: Function 'truncated' is not allowed in @bugs ...
```

The string parser rewrites BUGS `T(l, u)` and `C(l, u)` to `truncated(...)` and `censored(...)`, but neither name was in `BUGS_ALLOWED_FUNCTIONS`. The test suite only passed because `test/runtests.jl` registered both with `@bugs_primitive`.

**Change:** add `:truncated` and `:censored` to the default allowlist in `__init__`; drop the manual registration in `runtests.jl` so tests exercise the default; add a test that compiles one `T()` and one `C()` string program.

## 2. `dweib(a, b)` used the wrong scale

BUGS defines `dweib(a, b)` by the density `a·b·x^(a-1)·exp(-b·x^a)` (this formula is also in the docstring). Matching it to `Distributions.Weibull(shape, scale)` gives `scale = b^(-1/a)`. The code used `Weibull(a, 1/b)`, which agrees only when `a = 1`:

```julia
pdf(dweib(2.0, 3.0), 0.7)        # main: 0.1532
2*3*0.7^(2-1)*exp(-3*0.7^2)      # BUGS: 0.9657
```

**Change:** `Weibull(a, b^(-1 / a))`, docstring explains the mapping, test checks the density against the BUGS formula over several `(a, b, x)` and the `a = 1` exponential special case. Note this changes the posterior of the shipped Mice and Kidney examples (Weibull likelihood with estimated shape `r`); they previously did not match the BUGS parameterization.

## 3. `x ~ dt(μ, τ, ν)` with μ ≠ 0 or τ ≠ 1 could not compile without inits

```julia
compile(@bugs(begin x ~ dt(1, 4, 3) end), (;))
# ERROR: MethodError: no method matching iterate(::TDistShiftedScaled)
```

`dt` returns the package's own `TDistShiftedScaled` in that case, which implemented only `pdf`/`logpdf`. `BUGSModel` construction calls `Bijectors.bijector(dist)` (needs `minimum`/`maximum`) and `rand(dist)` to initialize the parameter.

**Change:** implement `minimum`, `maximum`, and `rand(rng, d)` (`μ + σ·rand(TDist(ν))`). Test samples it, checks location/scale, and compiles the model above.

## Tests

`Pkg.test(test_args=["elementary", "frontend", "compilation_source_gen"])` passes locally (includes doctests and the source-gen checks on Mice/Kidney/Magnesium).

### DIFF
--- a/JuliaBUGS/History.md
+++ b/JuliaBUGS/History.md
@@ -1,6 +1,6 @@
 # JuliaBUGS Changelog
 
-## Unreleased
+## 0.16.1
 
 ### Bug Fixes
 
--- a/JuliaBUGS/History.md
+++ b/JuliaBUGS/History.md
@@ -1,5 +1,13 @@
 # JuliaBUGS Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+- BUGS programs using truncation or censoring, `x ~ dnorm(0, 1)T(0, )` or `C(, c)`, compile out of the box. The string parser rewrites these to `truncated(...)` / `censored(...)`, which were missing from the default function allowlist, so `compile` rejected them unless the user registered both with `@bugs_primitive` first.
+- `dweib(a, b)` follows the BUGS parameterization, density `a b x^(a-1) exp(-b x^a)`, i.e. `Weibull(a, b^(-1/a))`. It previously built `Weibull(a, 1/b)`, which agrees only when `a = 1`. Posteriors of models with a Weibull likelihood and shape other than one (Mice, Kidney) change accordingly.
+- `x ~ dt(μ, τ, ν)` with a non-standard location or scale compiles without initial values: `TDistShiftedScaled` now implements `rand`, `minimum`, and `maximum`, which model construction and the unconstraining transform need.
+
 ## 0.16.0
 
 ### Highlights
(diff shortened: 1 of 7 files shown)


shravanngoswamii:
@sunxd3 Asking this question without looking at the PR: Does this PR also need changes in DoodlePPL code generation side? Because I remember playing with truncated() and censored()  handling while I was writing BUGS code generation...

sunxd3:
I'm not sure - sorry, could you verify? (Honestly I don't see how it could, but)

shravanngoswamii:
> I'm not sure - sorry, could you verify? (Honestly I don't see how it could, but)

If this PR changes any behavior than I guess change will be required. I just asked that question on phone when I saw truncated() and censored() are mentioned in PR. Never mind, I will have a careful look.

PR #536, opened 2026-09-03
Settle the VNChain bundle_samples ambiguity with DynamicPPL loaded

shravanngoswamii:
Sampling any model with `chain_type = VNChain` fails on 0.17.0 as soon as DynamicPPL is loaded, which every session using JuliaBUGS alongside Turing does. Every sampler is affected, NUTS included.

`Type{VNChain}` is a concrete singleton, so it beats `Type{<:FlexiChain{<:VarName}}` on specificity. A method has to match the alias exactly to win against the DynamicPPL extension.


sunxd3 (approved):
looks sensible

Issue #530, opened 2026-09-02
compiler: replace Core.eval-generated model functions


yebai:
JuliaBUGS constructs node functions with `Core.eval` when a model-building function is called. Evaluating the resulting model within the same ordinary function then fails under Julia's world-age rules:

```julia
using JuliaBUGS, Distributions, LogDensityProblems

@model function standard_normal((; x))
    x ~ Normal(0, 1)
end

function reproduce()
    model = standard_normal((;))
    return LogDensityProblems.logdensity(model, [0.0])
end

reproduce()
```

This raises a `MethodError` reporting that the matching node method is too new. The same cause affects AD preparation, gradient-free sampling, serialization, and other function-scoped operations. Several paths consequently depend on `Base.invokelatest`.

Replace runtime-generated methods with a world-age-safe representation. Candidates include a typed evaluator IR composed of callable objects and `Core.OpaqueClosure`. The latter may provide compiled callables without another dependency, but construction from runtime expressions uses compiler-facing APIs; stability across supported Julia versions and compatibility with AD require evaluation.

The solution should support `@model`, `@bugs`, parsed strings, and runtime `Expr` inputs through one compilation path. It must preserve caller-module name resolution, serialization, distributed execution, and evaluation performance.

Regression tests should cover direct log-density evaluation and gradient-based and gradient-free sampling from ordinary functions without `invokelatest`. Existing latest-world workarounds can then be removed.

PR #537, opened 2026-09-03
compiler: explore world-age-safe evaluation with MistyClosure

yebai:
JuliaBUGS creates node and log-density methods with `Core.eval`. When compilation and evaluation occur in the same function, those methods belong to a newer Julia world and cannot be called directly. This addresses #530.

This PR explores representing generated code as callable IR using `MistyClosure`. A small adapter specializes closures for the evaluation environment and permits the existing `invokelatest` workarounds to be removed. ForwardDiff, ReverseDiff, and Mooncake forward and reverse modes remain functional. Mooncake reverse mode requires a narrow extension because it does not currently support `MistyClosure` adjoints.

Representative primal measurements (`main` → PR):

- Graph: 1.99 → 2.04 µs; allocations unchanged at 2,112 B.
- Generated: 79 → 69 ns; 64 → 0 B.
- Twenty compilations: 105 → 28 ms; 15.3 → 0.76 MB.

Gradient timings show unresolved regressions:

- ForwardDiff graph: 2.51 → 5.00 µs.
- ReverseDiff graph: 36.3 → 40.7 µs.
- Mooncake graph: 131 → 156 µs.
- Mooncake generated: 0.37 → 13.5 µs.
- MooncakeForward generated: 2.64 → 3.46 µs.

This is opened for discussion. A better MistyClosure/OpaqueClosure implementation may exist and could replace this PR. I do not plan further work on this implementation.

EDIT: The revised MistyClosure adapter specializes calls by argument type and lets Mooncake differentiate retained source before native optimization, removing runtime fallback lookups and redundant AD machinery.

Preliminary Julia 1.12 gradient timings (`main` → original PR → revised):

- ForwardDiff graph: 2.51 → 5.00 → 2.82 µs.
- ReverseDiff graph: 36.3 → 40.7 → 35.7 µs.
- Mooncake graph: 131 → 156 → 125 µs.
- Mooncake generated: 0.37 → 13.5 → 0.32 µs.
- MooncakeForward generated: 2.64 → 3.46 → 2.60 µs.


sunxd3:
Some issues that the agents founds for note:

---

## 1. Graph-mode Mooncake gradient fails for any model compiled before `using Mooncake`

The whole suite loads Mooncake up front ([`test/runtests.jl#L45`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/test/runtests.jl#L45)), so no test has a node function whose retained world predates Mooncake. The `BUGSModelWithGradient` docstring shows the opposite order.

```julia
using JuliaBUGS, ADTypes, LogDensityProblems

model_def = @bugs begin
    μ ~ dnorm(0, 0.01)
    τ ~ dgamma(2, 2)
    for i in 1:N
        y[i] ~ dnorm(μ, τ)
    end
end
data = (N=10, y=collect(range(-1.0, 1.0; length=10)))
model = compile(model_def, data, (μ=0.0, τ=1.0))
x = JuliaBUGS.getparams(model)

using Mooncake

grad_model = JuliaBUGS.BUGSModelWithGradient(model, AutoMooncake(; config=nothing))
println(LogDensityProblems.logdensity_and_gradient(grad_model, x))
```

This branch:

```
ERROR: MethodError: no method matching rrule!!(::Mooncake.CoDual{Tuple{}, Mooncake.NoFData}, ::Mooncake.CoDual{@NamedTuple{N::Int64, μ::Float64, y::Vector{Float64}, τ::Float64}, ...}, ::Mooncake.CoDual{@NamedTuple{}, Mooncake.NoFData})
The function `rrule!!` exists, but no method is defined for this combination of argument types.
```

`main` prints `(-15.061651634162592, [2.962962962962963, -3.3306690738754696e-16])`.

Cause: [`ext/JuliaBUGSMooncakeExt.jl#L26`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/ext/JuliaBUGSMooncakeExt.jl#L26) pins the interpreter to `entry.world`. Mooncake's `is_primitive` runs `_is_primitive` through `invoke_in_world` at that world, where none of Mooncake's methods exist. The resulting `MethodError` is caught and treated as "this is a primitive" (`src/interpreter/contexts.jl`, the `e isa MethodError ? true : rethrow(e)` branch), so the entire closure body is handed to the fallback `rrule!!`, which has no method for it. The `CoDual{Tuple{}}` first argument in the error is the closure's empty captures. Any fix has to infer at a world no older than the extension's load world, e.g. `max(entry.world, world_when_ext_loaded)`; that keeps primal and derivative consistent unless a helper was redefined between `compile` and `using Mooncake`.

## 2. A package that compiles a model at top level loads, then throws on first use

No test precompiles a package that holds a model. (`eval_module=@__MODULE__` is needed on `main` too, so that part is not new.)

```julia
using Pkg, JuliaBUGS

pkg = joinpath(mktempdir(), "ModelHoldingPkg")
mkpath(joinpath(pkg, "src"))
write(joinpath(pkg, "Project.toml"), """
name = "ModelHoldingPkg"
uuid = "0f1e9a35-2d4b-4f5e-9c8a-7b3d2e1f0a11"
version = "0.1.0"

[deps]
JuliaBUGS = "ba9fb4c0-828e-4473-b6a1-cd2560fee5bf"
""")
write(joinpath(pkg, "src", "ModelHoldingPkg.jl"), """
module ModelHoldingPkg
using JuliaBUGS, JuliaBUGS.BUGSPrimitives
const model_def = @bugs begin
    μ ~ dnorm(0, 0.01)
    for i in 1:N
        y[i] ~ dnorm(μ, 1)
    end
end
const model = compile(model_def, (N=3, y=[0.1, -0.2, 0.3]), (μ=0.0,); eval_module=@__MODULE__)
end
""")

env = mktempdir()
Pkg.activate(env)
Pkg.develop(path=pkgdir(JuliaBUGS))
Pkg.develop(path=pkg)
Pkg.add("LogDensityProblems")
Pkg.precompile()

script = """
using ModelHoldingPkg, LogDensityProblems
println("log density: ", LogDensityProblems.logdensity(ModelHoldingPkg.model, [0.0]))
"""
run(`$(Base.julia_cmd()) --project=$env -e $script`)
```

This branch, in the fresh process:

```
ERROR: KeyError: key Symbol("##misty_closure#277") not found
  [1] getindex(h::Dict{Symbol, JuliaBUGS._MistyClosureEntry}, key::Symbol)
  [4] _misty_closure_entry   @ src/misty_closure.jl:23
  [5] macro expansion        @ src/misty_closure.jl:65   (the @generated body)
  [8] getparams              @ src/model/bugsmodel.jl:697
```

`main` prints `log density: -6.048339225812738`.

Three separate things go wrong here, and a fix needs to cover all of them:

- The registry at [`src/misty_closure.jl#L8`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/src/misty_closure.jl#L8) is a global inside JuliaBUGS. The downstream package fills it during its own precompilation, but mutations of another package's globals are not written into the downstream image, so the entry is gone in the new session while the `_MistyClosureFunction{ID}` object survives.
- The generator embeds a `_MistyClosureSpecialization` literal carrying `entry.world` into the cached code ([`#L80`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/src/misty_closure.jl#L80)). A world number from the precompile process is meaningless in a new process, so even a cached specialization would hand a bogus world to `invoke_in_world` at [`#L43`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/src/misty_closure.jl#L43).
- If the closure was already initialised during precompilation, the `OpaqueClosure` itself ends up in the image. That loads and then segfaults on the first call, independent of JuliaBUGS:

  <details><summary>opaque closure in a package image (click to expand)</summary>

  ```julia
  using Pkg
  pkg = joinpath(mktempdir(), "OCPkg"); mkpath(joinpath(pkg, "src"))
  write(joinpath(pkg, "Project.toml"), "name = \"OCPkg\"\nuuid = \"7c0d3a1e-1111-4222-8333-444455556666\"\nversion = \"0.1.0\"\n")
  write(joinpath(pkg, "src", "OCPkg.jl"), """
  module OCPkg
  const oc = Core.eval(@__MODULE__, :(Base.Experimental.@opaque (x::Int) -> x + 1))
  end
  """)
  env = mktempdir(); Pkg.activate(env); Pkg.develop(path=pkg); Pkg.precompile()
  run(`$(Base.julia_cmd()) --project=$env -e 'using OCPkg; println("loaded, oc.world = ", OCPkg.oc.world, ", current world = ", Base.get_world_counter()); println(OCPkg.oc(1))'`)
  ```

  ```
  loaded, oc.world = 38688, current world = 38687

  [1232685] signal 11 (1): Segmentation fault
  ```

  </details>

  So closure construction must be skipped whenever `ccall(:jl_generating_output, Cint, ()) == 1`.

## 3. Closure IDs are not unique across processes

[`src/misty_closure.jl#L94`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/src/misty_closure.jl#L94) uses `gensym`, a per-process counter. Two processes compiling two different models:

```julia
for body in ("x ~ dnorm(0, 1)", "x ~ dgamma(2, 2)")
    script = """
    using JuliaBUGS
    compile(@bugs(begin; $body; end), (;), (x=1.0,))
    println("$body  =>  ", only(keys(JuliaBUGS._MISTY_CLOSURE_ENTRIES)))
    """
    run(`$(Base.julia_cmd()) --project=$(Base.active_project()) -e $script`)
end
```

```
x ~ dnorm(0, 1)  =>  ##misty_closure#277
x ~ dgamma(2, 2)  =>  ##misty_closure#277
```

Today this is masked by item 2, because nothing persists. As soon as registry entries are made to survive package images, two packages can assign the same ID to different node functions and then share one `_MistyClosureFunction{ID}` type, one generated specialisation, and one Mooncake rule-cache entry. The wrong node function would run with no error. Any fix for item 2 needs process-unique IDs first.

## 4. A compiled model silently ignores later redefinitions

This is intended and tested ([`test/misty_closure.jl#L28`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/test/misty_closure.jl#L28)), so I list it for visibility, not as a bug. It is a user-visible behaviour change with no error, and the changelog entry sits under "Bug Fixes".

```julia
using JuliaBUGS, LogDensityProblems

prec_scale(x) = 1.0
JuliaBUGS.@bugs_primitive prec_scale
model_def = @bugs begin
    μ ~ dnorm(0, prec_scale(1))
end
model = compile(model_def, (;), (μ=0.5,))
println("before redefinition: ", LogDensityProblems.logdensity(model, [0.5]))

prec_scale(x) = 100.0
println("after redefinition:  ", LogDensityProblems.logdensity(model, [0.5]))
```

| | before | after |
|---|---|---|
| this branch | -1.0439385332046727 | -1.0439385332046727 |
| `main` | -1.0439385332046727 | -11.116353440210627 |

The generated log-density function is frozen the same way. In a Revise or REPL workflow, fixing a helper and re-evaluating gives stale numbers with nothing to indicate it.

## Notes, not failures

- **The compile cache is keyed on the exact world counter** ([`#L85`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/src/misty_closure.jl#L85)). Compiling `seeds` three times inside one loop keeps the registry at 6 entries. Doing it at top level as `m1 = compile(...)`, `m2 = compile(...)`, `m3 = compile(...)` gives 6, 12, 18, because on Julia 1.12 assigning a new global advances the counter. So in interactive use the cache rarely hits, each miss adds a type, a generated specialisation, and native code, and nothing is ever removed. `main` leaks one method per node per compile too, so this is a design property rather than a regression. Each miss also leaves the node's retained world non-current, for which Mooncake serves a fresh uncached interpreter (`get_interpreter(mode, world)`), so node rules are derived outside the shared cache. Measured prepare times were still within noise of `main`.
- **Inference runs inside the generator.** `Core.Compiler.typeinf_ircode` is called at [`#L15`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/src/misty_closure.jl#L15) from the `@generated` function at [`#L62`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/src/misty_closure.jl#L62). It is non-public and already needed the `stmt`/`inst` branch at [`#L68`](https://github.com/TuringLang/JuliaBUGS.jl/blob/913de2a3ff2aceaa4e32c8c15ebe459cc6dd5901/JuliaBUGS/src/misty_closure.jl#L68). Nothing failed in my runs; this is where the next Julia minor is most likely to bite.
- **The extension uses `rule_type`, `LazyDerivedRule`, `LazyFRule`, and `MooncakeInterpreter(; world)`**, none of which are public API. With the `0.5.33` compat entry, a Mooncake patch that renames any of them breaks every JuliaBUGS gradient.

For completeness: I did not find a time-to-first-gradient regression. Mooncake prepare times were within noise of `main` and gradient throughput was at parity. Primal graph-mode evaluation was roughly 10% slower from the extra closure indirection.

yebai:
Thanks, Xianda. Could we separate **what the model computes** from **the compiled functions used to evaluate it**?

Assume standard BUGS primitives—or custom primitives with stable meanings—with no hidden global constants or mutable state. Model-specific values come from literals, data, or parameters. Restoration requires a compatible environment; unrestricted `@model` helpers would need a separate contract.

Under these assumptions, the model’s expressions, structure, and explicit state define a fixed program. We can generate different functions to evaluate or differentiate that program without changing its meaning.

For example, `y ~ dnorm(μ, 1)` with observed `y = 2` should retain its density and gradient at a given `μ` across evaluation modes and restoration, within numerical tolerance.

This suggests:

- Building AD-compatible functions after the backend loads, using the unchanged model.
- Saving the model description and state, then rebuilding compiled functions when loaded.
- Ensuring package precompilation retains no runtime opaque closures or world numbers, including when it evaluates nodes.
- Using model-owned references instead of global closure IDs.
- Moving inference and closure construction outside generated functions.

Would this be a reasonable direction? It avoids freezing Julia’s entire method table, while the concrete implementation still needs validation.

yebai:
See, https://github.com/TuringLang/JuliaBUGS.jl/pull/541#issue-5478210455

Closed in favour of https://github.com/TuringLang/JuliaBUGS.jl/pull/541.

PR #538, opened 2026-09-04
Redirect the old DoodleBUGS path to DoodlePPL

shravanngoswamii:
`/JuliaBUGS.jl/DoodleBUGS/` still serves the pre-rename Vue app, because the publish step only ever writes `DoodlePPL/`. Every share link made before the rename points there, and until recently that was the only deployment able to open one, so the path cannot simply be deleted.

The path now serves a redirect that carries the query string and fragment across, which is what an old share link needs. A second deploy step with `clean` replaces the stale 2.3 MB build with that one page.

Draft until #535 raises the `doodleppl` pin. DoodlePPL is on `doodleppl@0.6`, which cannot read a share link, so redirecting before that lands would send old links somewhere strictly worse than where they go today.

Verified in Chrome: `/JuliaBUGS.jl/DoodleBUGS/?share=gz_ABC123#frag` lands on `/JuliaBUGS.jl/DoodlePPL/` with both the query and the fragment intact.

### DIFF
--- a/.github/workflows/DoodlePPLpublish.yml
+++ b/.github/workflows/DoodlePPLpublish.yml
@@ -6,6 +6,7 @@ on:
       - main
     paths:
       - 'DoodlePPL/**'
+      - 'DoodleBUGS/**'
       - '.github/workflows/DoodlePPLpublish.yml'
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


shravanngoswamii:
Doing this as a one-time change on gh-pages instead. The publish step scopes its clean to DoodlePPL, so a redirect placed there is not overwritten and the workflow needs no change.

PR #541, opened 2026-09-16
Evaluation with `OpaqueClosure`. 

### LINKED ISSUE #530: compiler: replace Core.eval-generated model functions

yebai:
Fix late AD loading and package precompilation/restoration in `misty`, retaining MistyClosure execution and JuliaBUGS's namespace for generated evaluation.

Each callable retains an ordinary Julia function as `f.source`, conceptually:

```julia
source = (evaluation_env, parameters) -> begin
    # Compute the model's log density.
end
```

This is a function object—not source text, typed SSA IR, or a Julia `@generated` function. Runtime inference produces typed SSA IR for each argument signature and builds a cached MistyClosure. Mooncake differentiates `f.source` through an overlay using available AD rules, allowing backend loading after model compilation.

Saving retains reconstruction information, not executable closures or the original world age. Loading reconstructs functions; specialisations capture the world available when prepared. Loading gradient-enabled models also rebuilds gradient preparation, requiring the backend beforehand.

Package-level models must own their source: define `@model` locally or use `eval_module=@__MODULE__` with primitives imported. Precompilation uses ordinary functions; opaque closures and AD caches stay outside package images. Prepare gradients after package loading.

Primitives and helpers must be **pure and retain stable semantics**: same inputs, same results, no observable side effects or hidden mutable state. 

Like `main`, this does not snapshot helpers. Cached specialisations may retain old definitions while new preparations observe edits. After editing helpers, recompile with `compile(model_def, data)` and prepare gradients again.

Warm benchmarks remain comparable to `misty`; generated density evaluation remains allocation-free.

Closes https://github.com/TuringLang/JuliaBUGS.jl/issues/530


sunxd3 (approved):
Thanks, Hong. This makes a lot of sense. No bone to pick.

sunxd3:
@shravanngoswamii could you give this a pass too?

shravanngoswamii:
This is pretty cool, and need a careful look, so I am taking my time.

yebai:
a66595ed is a small simplification: we can use Julia’s built-in `Core.OpaqueClosure` directly and drop the extra `MistyClosures` dependency. The existing source closure still handles AD, serialization, and respecialization, while opaque closures are created lazily for each signature and current world. Same behavior and lifecycle guarantees, with less machinery to maintain.

sunxd3:
On second review, I think there are some more decision points we can discuss.

https://github.com/TuringLang/JuliaBUGS.jl/issues/530 is an annoyance, but if `compile` and `sample` are both called from top level, `invokelatest` should not be needed.

e.g.

```julia
@model function standard_normal((; x))
    x ~ Normal(0, 1)
end

model = standard_normal((;))
LogDensityProblems.logdensity(model, [0.0])
```

should work when executed as a file or via cli.

---

I think the biggest interface regression with OpaqueClosure (OC) is that after a node function is redefined, the model needs a recompile, and until then it can give silent wrong results. Forgetting `@bugs_primitive` also needs a recompile.

The doc stated this, but it's not checked. One idea to check against it: the `eval`ed function is still retained, and although it can't be called because of world age, we can use its compiled code as a reference. Julia invalidates that code when a helper it depends on is redefined, so we know when the OC is stale. This will involve using Julia internals, but it seems tractable.

---

Also, this PR seems to regress performance in the case where `compile` and `sample` are called in the same function:

```julia
using JuliaBUGS, AbstractMCMC, AdvancedMH, Distributions, LinearAlgebra

model_def = @bugs begin
    for i in 1:N
        y[i] ~ dnorm(mu, tau)
    end
    mu ~ dnorm(0, 0.001)
    tau ~ dgamma(0.01, 0.01)
end

function fit(y)
    model = compile(model_def, (; N=length(y), y))
    sampler = RWMH(MvNormal(zeros(2), 0.01I))
    sample(model, sampler, 10_000; progress=false)
end

y = randn(200)
fit(y)        # warm up
@time fit(y)
```

| | main | PR |
|---|---|---|
| Julia 1.10 | 0.53 s, 383 MiB | 1.98 s, 761 MiB |
| Julia 1.12 | 0.52 s, 378 MiB | 2.96 s, 835 MiB |

This includes `compile` (~0.1 s on main, ~0.35 s on the PR), so most of the gap is sampling.

The reason seems to be that when `fit` starts, the model's functions don't exist yet. `compile` creates them mid-function, and `fit` can't see them. The wrapper asks for the node function's return type from `fit`'s world, gets no answer, and falls back to an untyped path on every node call. Main avoided this with one `invokelatest` inside `sample`, so this case was already fast on main. In a way, this PR avoids `invokelatest` but pays a price.

---

`compile` can be slower (rats: 0.38 → 0.77 s on 1.10, 0.43 → 1.28 s on 1.12), which is expected and tolerable IMO, but worth a mention.

yebai:
> I think the biggest interface regression with OpaqueClosure (OC) is that after a node function is redefined, the model needs a recompile, and until then it can give silent wrong results. Forgetting @bugs_primitive also needs a recompile.
> The doc stated this, but it's not checked. One idea to check against it: the evaled function is still retained, and although it can't be called because of world age, we can use its compiled code as a reference. Julia invalidates that code when a helper it depends on is redefined, so we know when the OC is stale. This will involve using Julia internals, but it seems tractable.

It may depend on perspective on JuliaBUGS, if one views it as a compiler for BUGS implemented in Julia. It is not surprising if we modify the primitives and requires running `compile` before the next `sample` call. The alternative perspective is that this is a Julia macro (just like DynamicPPL); then it would be a bit surprising that models can't automatically pick up new functions available. In my view, the former perspective is coherent and intuitive, and likely easier to internalise, as each `compile` gives back a compiled model which lives in a frozen world. 

> this PR seems to regress performance in the case where compile and sample are called in the same function:

That is something we could investigate a bit further. Might be better for a separate PR so this one is focused.

sunxd3:
sounds good, thanks for working on this!

sunxd3:
@shravanngoswamii please don't hesitate to comment if you got time to do a post merge review

PR #543, opened 2026-09-17
Store the BUGS examples as files under BUGSExamples/ and give every example a docs page

shravanngoswamii:
An installed JuliaBUGS has no checkout, so `JuliaBUGS.BUGSExamples` reads a snapshot of `BUGSExamples/` shipped as a lazy artifact, and a checkout reads the directory itself. The parsed `model.bugs` files compile to the same models as the hand-written Julia versions they replace, except that Asia, Eye Tracking, and Hepatitis keep the published dotted names the Julia versions had rewritten, so Asia's reference results are keyed `lung.cancer`. Methadone's BUGS text is written from the Julia model, since that file never carried it. The example pages fetch their published runs from the mcmcjs bundle index at view time, and the widget version and mcmcjs URLs live only in `docs/src/assets/examples.js`. Merging triggers the snapshot workflow, which creates the artifact release on its own.


shravanngoswamii:
@yebai @sunxd3 Any thoughts on this approach? If it looks good, I will clean it more and we are ready to merge.

sunxd3:
what is the biggest problem this solves?

shravanngoswamii:
> what is the biggest problem this solves?

Before this, each example existed three times, a hand-translated `@bugs` block, the original BUGS text beside it, and the model and data typed again on its docs page, and the copies had drifted. Now the BUGS program and its data are files, read by the package, by every docs page, and by mcmcjs, whose mcmc run takes an example folder as it is and publishes the runs the pages show. It is also what let Volume 3 and Methadone in, since nothing is baked into the module at load time any more.

And, I guess we wanted to separate examples from package.

sunxd3 (approved):
LG as far as I can tell

sunxd3:
thanks!

